PR #4, opened 2022-06-07
Change physics to bevy_rapier2d

odecay:
I've done a replacement of the heron physics crate and the physics code using it for bevy_rapier2d. I tried to stick closely to the original implementations when possible with the equivalent rapier API, with some small changes where necessary.

I believe functionality continues to be the same as it was.

This rewrite was motivated by changes in the bevy_rapier crate which make it more ergonomic and should lead to better support and keeping pace closely with rapier proper since the bevy_rapier is an official integration.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -110,12 +110,6 @@ dependencies = [
 
 [[package]]
 name = "arrayvec"
-version = "0.5.2"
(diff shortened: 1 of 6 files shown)


edgarssilva:
I have taken a quick look and I really like the simplifications, they look great.
I'll merge it when I get home.

PR #14, opened 2022-06-27
BodyLayers Enum More Readable

zicklag:
Using bitshift syntax makes these kinds of enums so much easier to read. :)


odecay:
im not 100% sure I agree. I am unfamiliar with that syntax but I can understand it after looking it up. How would it work when setting bitmask manually like in https://github.com/fishfight/punchy/blob/9e20b3b4d010f1f0d2622095ef0c1a9eb61843f7/src/main.rs#L218

My other reasoning for using current syntax is that its whats used in the [Rapier Docs](https://rapier.rs/docs/user_guides/bevy_plugin/colliders#collision-groups-and-solver-groups) so if someone is trying to follow along there to figure out what they are doing this is what they would encounter.

Tangential to that it might be good to add more enum variants for "All" of 0b1111 and use those in the places 0b1111 is currently used.

odecay:
Okay looks good. Will you rebase? I added some bundles for setting up components so the section in main where 0b1111 was being used changed.

odecay:
awesome I was going to ask about not having to cast as u32 and that works great.

PR #15, opened 2022-06-27
Add Optimization Flags to Cargo.toml

zicklag:
Feel free to tweak anything you don't like. :)


odecay:
Oh this is good, I have been using a similar config in a different project but with [profile.dev] opt-level = 1. Probably good to use some optimization in dev.

zicklag:
Yeah, you're probably right. Updated opt-level = 1

PR #16, opened 2022-06-28
Add GitHub actions CI, with format check

64kramsystem:
This blocks code that is not formatted (note: Actions may need to be activated on this repo).

Note that the other action that the README currently suggests:

```sh
cargo clippy -- -W clippy::correctness -D warnings
```

is not currently possible (without tweaks) because of some Bevy issues/characteristics, e.g. the following (known) problem:

    error: call to `std::mem::forget` with a value that does not implement `Drop`. Forgetting such a type is the same as dropping it
      --> src/main.rs:102:10
        |
    102 | #[derive(Bundle)]
        |          ^^^^^^
        |
    note: argument has type `Enemy`
      --> src/main.rs:102:10
        |
    102 | #[derive(Bundle)]
        |          ^^^^^^
        = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#forget_non_drop
        = note: this error originates in the derive macro `Bundle` (in Nightly builds, run with -Z macro-backtrace for more info)

Some of the suggestions are worth applying though, e.g. unnecessary `clone()` calls.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,15 @@
+name: CI
+
+on:
+  pull_request:
+
+jobs:
+  check_formatting:
+    runs-on: ubuntu-latest
+    name: Check Rust formatting
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/cargo@v1
+        with:
+          command: fmt
+          args: --all -- --check


odecay:
It seems worthwhile to have CI setup. 
I know it possible to work around the bundle issue in bevy https://github.com/bevyengine/bevy/issues/4601#issuecomment-1114332185.
We could either delay setting up warnings till that is fixed or use the workaround.

64kramsystem:
> It seems worthwhile to have CI setup. I know it possible to work around the bundle issue in bevy [bevyengine/bevy#4601 (comment)](https://github.com/bevyengine/bevy/issues/4601#issuecomment-1114332185). We could either delay setting up warnings till that is fixed or use the workaround.

Definitely. Since that task has some complexity involved, I think it's more efficient to split the CI tasks in separate PRs:

- format (this)
- clippy
- build (some level of smoke testing, e.g. on multiple platforms)

odecay:
PR away, I was thinking for build we would base off https://github.com/bevyengine/bevy_github_ci_template/blob/main/.github/workflows/release.yaml

Would be nice to get that set up for next release. On top of that maybe we could set up as a github pages hosted wasm builds on tags/release.

64kramsystem:
> PR away, I was thinking for build we would base off https://github.com/bevyengine/bevy_github_ci_template/blob/main/.github/workflows/release.yaml
> 
> Would be nice to get that set up for next release. On top of that maybe we could set up as a github pages hosted wasm builds on tags/release.

Rust builds need to be carefully planned. AFAIK open source projects have 2000 minutes per month, which is not little, but build time in the Rust world is... something :wink: especially, when one builds for multiple platforms. Oh, and Bevy is B-I-G :smile:.

Like that wasn't enough, sometime seemingly little known in the Github Actions world is that caching is not shared across PRs. This means that if each PR will rebuild the project from scratch (on the first build).

This is unfortunately not a theoretical matter, since with the rusty game jam project, we (our team) have exhausted our credit :exploding_head:.

Anyway, nothing impossible, in particular, if a solid solution is found, it's something worth upstreaming. I'll follow up on Discord and/or here. :wave:

PR #19, opened 2022-06-28
Camera Zoom Fix

Byteron:
Prevents Camera from being negative. Also uses delta time for zoom in, so frame rate doesn't matter.


Byteron:
for reference, that's what happened previously when zooming in too far:
image
the scale became negative and thus the rendering inverted.

edgarssilva:
Looks good to me

PR #20, opened 2022-06-28
Camera Panning Fix

Byteron:
Currently, when panning, the camera doesn't move from the camera, however the parallax background layers are moving, the further back the faster.

This PR fixes / improves the camera panning.
The Panning is now handled with a separate component storing the offset from the player. this way you can pan the camera independently from the player position. If you move the player, the camera keeps its offset.

Also, the parallax backgrounds moves correctly with the camera panning.

unknown_2022 06 28-15 15

### DIFF
--- a/src/camera.rs
+++ b/src/camera.rs
@@ -1,9 +1,10 @@
 use bevy::{
     core::Time,
     input::Input,
+    math::Vec2,
     prelude::{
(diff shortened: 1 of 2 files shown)


Byteron:
also for comparison, this is how it is in master
unknown_2022 06 28-16 22
you see, up and down sort of works, but left and right is not accurate. the camera stays on the player, but the background moves weirdly.

PR #22, opened 2022-06-29
Implement First Pass Scene Assets

zicklag:
This removes all the hard-coding for the scene spawning. Everything is loaded from assets.

I co-located the sprites for the fighters with the fighter YAML, but they could just as well be moved to a separate `sprites` folder if we wanted. It's setup so that when one asset references another asset the path is relative to the current file by default, but paths starting with a `/` are relative to the asset directory instead.

If we continue to put everything in assets, such as the UI theme images, etc. that are all referenced by a central `.game.yaml`, then the cool part is that the game becomes like a "punchy engine" where anybody can bring their own `.game.yaml` and totally change the feel and theme of the game.

This isn't implemented yet, but it's easy to do: you can even hook the web build of the game up so that you can use a query string to specify the asset URL to your own punchy game that you host on GitHub pages, making it a punchy game web player, and anybody can just host their assets on GitHub and then send people a link to play it with the punchy web player.


odecay:
Started looking it over. Looks good so far.
Could you explain the metadata?

src/camera.rs
```diff
@@ -65,7 +65,11 @@ pub fn camera_follow_player(
     mut camera_query: Query<(&mut Transform, &Panning), (With<Camera>, Without<Player>)>,
     mut move_event_writer: EventWriter<ParallaxMoveEvent>,
 ) {
-    let player = player_query.single().translation;
+    let player = if let Ok(player) = player_query.get_single() {
```
odecay, line 68: I agree with these checks using .get_single() instead of .single() if we are loading scenes or if players get despawned it would panic otherwise, but can this be shortened to
```rust
if let Ok(player) = player_query.get_single(){
//rest of system goes here
//I also advocate for using player.translation in place of everywhere player is used in this system
}

File: src/main.rs [hunk 2]
```diff
@@ -384,7 +352,11 @@ fn player_attack(
     keyboard: Res<Input<KeyCode>>,
     time: Res<Time>,
 ) {
-    let (mut state, mut transform, animation, facing) = query.single_mut();
+    let (mut state, mut transform, animation, facing) = if let Ok(player) = query.get_single_mut() {
```
odecay, line 355: could this be (similar nit to the one in camera.rs)
```rust
If let Ok(mut state, mut transform, animation, facing) = query.get_single_mut() {
//body of system goes inside here
}

File: src/movement.rs [hunk 3]
```diff
@@ -42,7 +42,12 @@ pub fn player_controller(
     keyboard: Res<Input<KeyCode>>,
     time: Res<Time>,
 ) {
-    let (mut state, stats, mut transform, facing_option) = query.single_mut();
+    let (mut state, stats, mut transform, facing_option) =
```
odecay, line 45: same as in main and camera
```rust
if let Ok(mut state, stats, mut transform, facing_options) = query.get_single_mut() {
//system body
}

File: src/camera.rs (hunk 1 above)
Comment (odecay) [reply src/camera.rs:68]: I could see a case for not doing it this way because of rightward drift but it seems more straightforward.

Also eventually we will want to come back and replace all the .get_single() for players when we start having more than 1 player although I imagine at least our camera handling will have to change at that point.

File: src/main.rs [hunk 4]
```diff
@@ -240,132 +210,130 @@ fn setup(
...
+            &Handle<Fighter>,
+            Option<&Player>,
+            Option<&Enemy>,
+        ),
+        Without<Stats>,
+    >,
+    fighter_assets: Res<Assets<Fighter>>,
+) {
+    for (entity, transform, fighter_handle, player, enemy) in fighters.iter() {
+        let fighter = if let Some(fighter) = fighter_assets.get(fighter_handle) {
```
odecay, line 299: same here with the shortening
```rust
if let Some(fighter) = fighter_assets.get(fighter_handle) {
//rest of load_fighters here
}

odecay (commented):
Okay looking really nice, On 2nd look I see what the metadata is.
Loading and spawning different entity types will be much simplified from what it was and the yaml files are nice and readable.

erlend-sh:
> the cool part is that the game becomes like a "punchy engine" where anybody can bring their own .game.yaml and totally change the feel and theme of the game.
That’s definitely the direction we wanna be going with. We also have plans to incorporate https://github.com/makspll/bevy_mod_scripting after Bevy v0.8

odecay (approved):

PR #30, opened 2022-06-29
Hot reload

zicklag:
Hot reload fighters and levels!

There is one other way we could have implemented this which has it's own trad-offs.

It's currently implemented by having a system watch for asset updated events. There is one such system for each kind of asset, in this case `Level`s and `Fighter`s. There is a little bit of duplication of the level/fighter spawn logic in the hot reload systems, because we are updating some of the same components, but not all of the components.

The alternative strategy for fighers ( though not for levels because it involves inserting resources ) would be to insert `Handle<Stats>` components onto the fighter entities instead of `Stats` components, for instance.

While this removes the need for a dedicated hot reload system and any duplication in the update logic, it also means that we can't just access the `Stats` component on a player without also adding `Res<Assets<Stats>>` to the system and making sure to do a `assets.get(stats_handle)`. So the systems accessing player stats becomes more a little more verbose.

I'm not sure what's going to be better in the long run, and we don't have to use one strategy for everything, we can take it on a case-by-case basis.

I feel like right now, because the hot reload systems are simple and small, it's probably not a big deal to have the small amount of duplication, but I could go either way with it if you want to use handles instead.

### DIFF
--- a/assets/beach/beach.level.yaml
+++ b/assets/beach/beach.level.yaml
@@ -1,3 +1,4 @@
+background_color: [101, 131, 162]
 parallax_background:
   layers:
     - speed: 0.98
--- a/src/main.rs
+++ b/src/main.rs
@@ -2,7 +2,10 @@
 #![allow(clippy::forget_non_drop)]
 
 use bevy::{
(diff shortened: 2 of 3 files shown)


odecay:
I think CI failed on this because I accidentally left a space or tab on blank line when I resolved conflicts in webui.

PR #32, opened 2022-06-30
Establish Web Build

zicklag:
Runs in a browser! Play it here: https://zicklag.github.io/punchy/demo/

punch-web

---

I added a `justfile` to help automate the build steps during development using [`just`](https://github.com/casey/just) but we can use plain shell scripts or something else if you want.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -30,6 +30,7 @@ mod consts;
 mod item;
 mod metadata;
 mod movement;
+mod platform;
 mod state;
 mod ui;
 mod y_sort;
@@ -186,6 +187,7 @@ fn main() {
             ..Default::default()
         })
         .add_event::<ThrowItemEvent>()
+        .add_plugin(platform::PlatformPlugin)
         .add_plugin(RapierPhysicsPlugin::<NoUserData>::pixels_per_meter(100.0))
         .add_plugin(AttackPlugin)
         .add_plugin(AnimationPlugin)
--- a/wasm_resources/index.html
+++ b/wasm_resources/index.html
@@ -1,0 +1,143 @@
+<html>
+  <head>
+    <meta charset="UTF-8" />
+    <style>
+      body {
+        margin: 0;
+        background-color: black;
+      }
+      canvas {
+        background-color: black;
+        /*
+          This is a workaround for high DPI screens which winit or something
+          has an issue with where it will resize the canvas to smaller than
+          the full width of the screen.
+        */
+        width: 100vw !important;
+        height: 100vh !important;
+      }
+    </style>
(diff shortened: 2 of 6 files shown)


src/main.rs
```diff
@@ -30,6 +30,7 @@ mod consts;
 mod item;
 mod metadata;
 mod movement;
+mod platform;
```
odecay, line 33: right now this does nothing outside of wasm build correct? Should it be under the same `#[cfg(target_arch = "wasm32")]` that you put the actual code in the plugin under?

src/main.rs
```diff
@@ -186,6 +187,7 @@ fn main() {
             ..Default::default()
         })
         .add_event::<ThrowItemEvent>()
+        .add_plugin(platform::PlatformPlugin)
```
odecay, line 190: same comment as above. Could move this to its own line like 
```rust
#[cfg(target_arch = "wasm32")]
app.add_plugin(platform::PlatformPlugin)

File: src/main.rs (hunk 1 above)
Comment (odecay) [reply src/main.rs:33]: I could see this being generally useful if we end up needing more platform specific stuff though in which case it would make sense to always leave it in.

File: wasm_resources/index.html [hunk 3]
```diff
@@ -0,0 +1,143 @@
...
+    </div>
+    <script>
+      // Insert hack to make sound autoplay on Chrome as soon as the user interacts with the tab:
+      // https://developers.google.com/web/updates/2018/11/web-audio-autoplay#moving-forward
+      (function () {
+        // An array of all contexts to resume on the page
+        const audioContextList = [];
+
+        // An array of various user interaction events we should listen for
+        const userInputEventNames = [
```
odecay, line 86: I don't really understand what this gets us but I'm assuming adding these input event listeners could be generally useful in the future?

odecay (commented):
Nice level of polish in this setup, with loading + canvas update on window resize handler!

src/main.rs
zicklag: Yeah, I figured out save data abstraction would go in there once we get to it, so it should always be installed, and we just put our platform specific stuff in there.

wasm_resources/index.html
zicklag: All of that is actually solely to fix audio contexts in Chrome and does nothing to impact the game otherwise.

On chrome, the web page isn't allowed to auto-play sounds before the user clicks on the screen or performs some other interaction with the page. what that code does is wait for some interaction and once it gets it, it resumes the audio context, which Chrome will allow because there has been a user interaction at that point.

src/main.rs
odecay: sounds good

wasm_resources/index.html
odecay: Okay seems fine :+1:

PR #36, opened 2022-07-01
Animation Loop

### LINKED ISSUE #35: Enemies stand up briefly before death
for one frame or so, the enemy is shown standing before dying.

Byteron:
This PR implements a `repeat` value for animations, that defines whether or not an animation should be played in a loop or not.

fixes #35


odecay:
Very nice, I like the change of adding looping to the animation in asset definition.

PR #39, opened 2022-07-02
Remove Warning on Compile

Byteron:
I'm not sure if this is necessary, but this PR removes the warning that is currently emitted when compiling.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -37,7 +37,7 @@ mod ui;
 mod y_sort;
 
 use animation::*;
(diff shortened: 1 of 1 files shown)


zicklag:
I ran into this same warning and realized it was because the `Attack` import is used when running with the `debug` feature. I fixed it as a part of #38.

odecay:
> just using the full path to `attack` so that we don't need the `use` statement.

this is my prefered way of fixing, gonna close this pr

PR #45, opened 2022-07-02
Allow Toggling Fullscreen With F11

zicklag:
Let me know if you want this in it's own module. I almost put it in a new module, but I wasn't sure what a good module to put it in was, so I figured I'd just ask what you thought first.

It almost could go in `platform`. :thinking:


edgarssilva:
A simple and nice feature :)

PR #47, opened 2022-07-03
Move move_in_arc and rotate systems to PostUpdate stage

### LINKED ISSUE #37: Throwable Item animation bug
https://user-images.githubusercontent.com/1379590/176952312-0ee9e997-d23e-4c85-9e9e-1e0e4dff3ab2.mp4

When throwing items with T, sometimes one will show up in front of the player toward the ground.
Seems to only happen when throwing multiple in succession but could also just be easier to notice that way.

64kramsystem:
This is required, in order to avoid 1-frame lag. With the previous design, they would not be able to find items spawned in the same frame.

Closes #37, however, it'd be much better to fully redesign the systems flow in order to take into account 1-frame-lag issues. Moving lag-sensitive systems into PostUpdate makes sense, but it's cleaner design to create one (or more) adhoc stages, in order to structure the flow clearly.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -234,10 +234,8 @@ fn main() {
                 .with_system(kill_entities)
                 .with_system(knockback_system)
                 .with_system(move_direction_system)
-                .with_system(move_in_arc_system)
                 .with_system(throw_item_system)
                 .with_system(item_attacks_enemy_collision)
-                .with_system(rotate_system)
                 .with_system(set_target_near_player)
                 .with_system(move_to_target)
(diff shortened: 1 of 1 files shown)


edgarssilva:
Nice one, we are looking to add system ordering and labeling, which should help fix this and future problems alike.

64kramsystem:
> Nice one, we are looking to add system ordering and labeling, which should help fix this and future problems alike.

You refer to something like `with_system(my_system.after(other_system))` (or the equivalent form using labels)?

Issue #48, opened 2022-07-03
Possible to get stuck at bottom of map


odecay:
It is possible to get stuck toward the bottom of the map such that it is only possible to move in X direction.
I vaguely remember it relating to the way the character is moved slightly downwards by the belly flop attack

https://user-images.githubusercontent.com/1379590/177055074-13eba39d-732c-4ceb-b8da-900c66b6bc8a.mp4

64kramsystem:
The belly flop doesn't play well with [this](https://github.com/fishfight/punchy/blob/2f7d84804876a2e832a95d52d55c1e0627a9e53f/src/movement.rs#L75-L76). Not sure what's the best way to fix it. I think that the belly flop should _not_ decrease the Y. I haven't played many beat-em ups, but the one I've played attacked only across the X axis.

64kramsystem:
Something really interesting is going on. It gets also stuck at the top, so the shift is relative to... something :laughing:

64kramsystem:
Ok. This is another case of the 1-frame-lag (or something similar). The first frame of the animation runs sometimes for 7 (Bevy) frames, sometimes for 8. When it runs for 7 frames, it causes the computation to shift the sprite up less than it should.

64kramsystem:
This diff displays the problem:

```diff
diff --git i/src/main.rs w/src/main.rs
index 54838af..fffb59f 100644
--- i/src/main.rs
+++ w/src/main.rs
@@ -628,6 +628,7 @@ fn unpause(keyboard: Res<Input<KeyCode>>, mut commands: Commands) {
 }
 
 fn player_attack(
+    mut frames_count: Local<u32>,
     mut query: Query<(&mut State, &mut Transform, &Animation, &Facing), With<Player>>,
     keyboard: Res<Input<KeyCode>>,
     time: Res<Time>,
@@ -649,10 +650,24 @@ fn player_attack(
                 }
             }
 
+            if animation.current_frame == 0 {
+                *frames_count += 1;
+                println!("FC0:{}", *frames_count);
+            }
+
+            if animation.current_frame < 1 {
+                transform.translation.y += 100. * time.delta_seconds();
+            } else if animation.current_frame < 2 {
+                transform.translation.y -= 100. * time.delta_seconds();
+            }
+
             if animation.current_frame < 1 {
-                transform.translation.y += 180. * time.delta_seconds();
-            } else if animation.current_frame < 3 {
-                transform.translation.y -= 90. * time.delta_seconds();
[...]

PR #49, opened 2022-07-04
Workaround for player getting stuck at the bottom

64kramsystem:
Very tricky one. It's currently unclear why the first frame may run for less Bevy frames than intended; possibly, this is Yet Another One-frame-lag Issue™.

Addresses #48 (unsure if it should close it).

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -631,6 +631,7 @@ fn player_attack(
     mut query: Query<(&mut State, &mut Transform, &Animation, &Facing), With<Player>>,
     keyboard: Res<Input<KeyCode>>,
     time: Res<Time>,
+    mut start_y: Local<Option<f32>>,
 ) {
     if let Ok((mut state, mut transform, animation, facing)) = query.get_single_mut() {
         if *state != State::Attacking {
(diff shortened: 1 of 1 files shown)


64kramsystem:
I've classified this as workaround, but even if animations were working as expected, I think it's still a solid approach to store the start Y coordinate and restore it.

64kramsystem:
> Will this style of storing data for a system in a local cause issues with multiple players? The beginnings of that work just got merged in so its now possible to have more than 1 player.

No, I think not :smile: 

An alternative approach (1) is just to clamp to Y range. This is the simplest :smile:  We do have boundaries, so it'd work.

Another approach (2) is to model each attack by one bundle/entity, and a dedicated system; in this case, we'd store in a component the start position.

Yet another approach (3) is to try to make the maths exact, but it'd still require to store some state in some form (time passed since start of the attack), so it'd be very similar in spirit to apprach 2.

If you have any preference, let me know :smile:

odecay:
2 would be my preferred solution, but I am okay with using as is right now if it works. I was wondering if both players attack, which Y will the get reset to in current implementation? If that is not an issue then this is good for now.

PR #52, opened 2022-07-06
Implement Input Mapping

zicklag:
Here's two players controlled both with gamepads, and keyboard controls work too!

punchy02-players

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ getrandom = { version = "0.2", features = ["js"] }
 bevy_mod_debugdump = { version = "0.4", optional = true }
 bevy-inspector-egui = { version = "0.11", optional = true }
 bevy-inspector-egui-rapier = { version = "0.4", optional = true, features = ["rapier2d"] }
+leafwing_input_manager = { git = "https://github.com/zicklag/leafwing-input-manager.git", rev = "45c40f3d" }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 web-sys = { version = "0.3", features=["Window"] }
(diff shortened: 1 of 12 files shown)


zicklag:
@odecay this is ready for review now.

It replaces _all_ game input with mapable actions in three categories: `player`, `camera`, and `menu`. There is no longer any occurrence of `KeyCode` anywhere in the code, it's all mapped and loaded from the `GameMeta` asset. :)

You can also play two players, now, though I believe there's a weird bug where the second player cannot move while the first player is flopping. I haven't investigated yet, but I think we can fix that after this PR.

I'm going to follow up with a PR to allow you to navigate the menus with the gamepad, and then after that I'll look into getting two players working without issue.

odecay:
> hough I believe there's a weird bug where the second player cannot move while the first player is flopping

Maybe its because of something I did in state checking for that system? Using return instead of break? https://github.com/fishfight/punchy/blob/0060a90bd2fed813a0bf5ae06593ec0b914a5ed1/src/main.rs#L664

odecay:
> it replaces all game input with mapable actions in three categories: player, camera, and menu. There is no longer any occurrence of KeyCode anywhere in the code, it's all mapped and loaded from the GameMeta asset. :)

Awesome! Review + merge will be my first priority after the v0.0.2 release is up today.

> I'm going to follow up with a PR to allow you to navigate the menus with the gamepad, and then after that I'll look into getting two players working without issue.

:+1:  Sounds good!

assets/default.game.yaml
```diff
@@ -12,6 +12,161 @@ main_menu:
     image: ui/main-menu-background.png
     image_size: [1919, 1027]
 
+default_input_maps:
```
odecay, line 15: Very nice how clean it is to have analogs + dpad + keyboard all share input maps

Cargo.toml
```diff
@@ -42,4 +43,7 @@ opt-level = 1
 lto = true
 codegen-units = 1 # Improves physics performance for release builds
 
-#TODO turn this into a debug feature
+[patch.crates-io]
+# Patch to enable improved DPad support. It's a merge of the following PR to Bevy 0.7.0 so we can
+# use it before 0.8 is release: https://github.com/bevyengine/bevy/pull/5220.
```
odecay, line 48: checked out the patch, looks relatively simple lets hope it gets merged :+1:

PR #56, opened 2022-07-08
Add Menu Gamepad Bindings

zicklag:
- Updates the leafwing-input-manager branch we are using to include
  latest API improvements. Check it out! Gamepad bindings are even cooler now!
- Updates buttons to have a visible focused state
- Allow controlling the menu focus with the gamepad.

It's now possible to fully navigate the game and menu without a mouse or keyboard, and you can tell when buttons are selected:

menu-highlight-buttons

---

I realized that since the commits I'm referencing for the leafwing-input-manager dependency are in my fork and not in the official repo on the master branch, once I delete my fork or branches, those commits will stop existing. It just breaks the ability to go back to any commit in history and build the game at that point.

Since we'll be hopefully getting the branch merged soon to master, where it will have a permanent commit ref, we may want to wait to merge this until that happens, unless you aren't really worried about a couple commits not having eternally reproducible builds.

PR #58, opened 2022-07-08
Move Game Loading To It's Own Module

zicklag:
Depends on #56 ( just to avoid having to resolve conflicts ).

Made an attempt to improve the understandability of the game loading by
not using a somewhat cryptic system chain and just using normal systems
and a custom system param.

Hopefully the system param isn't more confusing than the system chain I used earlier?

Let me know if this isn't actually helping.


zicklag:
Rebased this and it should be ready for merge.

odecay:
Bounds + system param are a bit complex but it still seems much nicer than previous.

PR #59, opened 2022-07-09
Localization & French Translation

zicklag:
Depends on #58.

Since it didn't take me long and it's better to do earlier so we don't have to re-do all our strings later, I setup localization!

Currently there's a French and English translation and it's super easy to add new translations. It will auto-detect the language from the users system, even in the browser, which saves us from making a language selection menu for now. :)

### DIFF
--- a/src/assets.rs
+++ b/src/assets.rs
@@ -23,13 +23,6 @@ pub fn register(app: &mut bevy::prelude::App) {
         .add_asset_loader(EguiFontLoader);
 }
 
-// An error that could ocurr during asset processing
-#[derive(thiserror::Error, Debug)]
-pub enum AssetLoaderError {
-    #[error("Could not parse YAML asset: {0}")]
-    DeserializationError(#[from] serde_yaml::Error),
-}
-
 /// Calculate an asset's full path relative to another asset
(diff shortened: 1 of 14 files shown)


zicklag:
This should be ready now. Once https://github.com/kgv/bevy_fluent/pull/21 is merged, we can update the `bevy_fluent` git dependency and we'll get hot reload for the translations.

zicklag:
Found that there is a translation selection issue on WASM actually, got to fix that first.

zicklag:
OK, fixed it. Should be good to go now.

src/assets.rs
```diff
@@ -23,13 +23,6 @@ pub fn register(app: &mut bevy::prelude::App) {
         .add_asset_loader(EguiFontLoader);
 }
 
-// An error that could ocurr during asset processing
-#[derive(thiserror::Error, Debug)]
-pub enum AssetLoaderError {
```
odecay, line 28: Just checking to make sure this is no longer necessary or useful..

zicklag: Yeah, it wasn't actually used. It was used in the code I had that I copied the old loader from, but it didn't end up being incorporated.

Issue #29, opened 2022-06-29
Multiple Players


odecay:
Prerequisite for actual multiplayer,
Systems which work on our current single entity with player component should be changed to work with multiple players.
Camera follow player system will have to change. Could be based centered on average point between all players to start.
Input will have to support multiple players, each should have their own input map.

- [ ] Camera refactor to support multiple players
- [ ] Input refactor to support multiple players

64kramsystem:
> Camera follow player system will have to change. Could be based centered on average point between all players to start.

In beat 'em ups, the camera typically moves when the players go out of an (invisible) box, with some extra rules (ie. boss fight, enemies wave etc.). I've started the groundwork for this in #62.

odecay:
I'll do part 3 of this tomorrow.

odecay:
Done in https://github.com/fishfight/punchy/commit/c36cecab45c44a5205d97291767fa3c900f07a15. Could probably still use improvement.

PR #62, opened 2022-07-10
Proper camera logic, and new player movement limitations

64kramsystem:
Closes the first task of #29; the camera has now a proper logic, typical of beat 'em ups.

Boss-related logic is not yet implemented, since there are no bosses yet :laughing:

### DIFF
--- a/assets/default.game.yaml
+++ b/assets/default.game.yaml
@@ -1,5 +1,8 @@
 start_level: beach/beach.level.yaml
 camera_height: 448
+# Allowed x distance between players, as a fraction of the screen width.
+# This is also used for camera move, so be careful.
+max_players_x_distance_fraction: 0.25
 
 main_menu:
   title_font:
--- a/src/movement.rs
+++ b/src/movement.rs
@@ -10,8 +10,13 @@ use bevy::{
 use leafwing_input_manager::prelude::ActionState;
 
 use crate::{
-    animation::Facing, consts::{self, LEFT_BOUNDARY_MAX_DISTANCE}, input::PlayerAction, item::ThrowItemEvent, metadata::GameMeta,
-    state::State, ArrivedEvent, DespawnMarker, Player, Stats,
+    animation::Facing,
+    consts::{self, LEFT_BOUNDARY_MAX_DISTANCE},
+    input::PlayerAction,
+    item::ThrowItemEvent,
+    metadata::GameMeta,
+    state::State,
+    ArrivedEvent, DespawnMarker, Player, Stats,
 };
 
 #[cfg_attr(feature = "debug", derive(bevy_inspector_egui::Inspectable))]
@@ -111,7 +116,7 @@ pub fn player_controller(
         })
         .collect::<Vec<_>>();
 
-    if player_dirs.len() == 2 {
+    if player_dirs.len() > 1 {
         let window_width = windows.get_primary().unwrap().width();
         let max_players_x_distance = window_width * game_meta.max_players_x_distance_fraction;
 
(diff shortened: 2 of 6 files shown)


64kramsystem:
Ok, this is ready for review etc. Note that this logic has severe limitations. This is intentional; fully implementing the new camera logic requires extensive changes, which IMO should not be applied in a single PR.

assets/default.game.yaml
```diff
@@ -1,5 +1,8 @@
 start_level: beach/beach.level.yaml
 camera_height: 448
+# Allowed x distance between players, as a fraction of the screen width.
+# This is also used for camera move, so be careful.
+max_players_x_pct_distance: 0.25
```
odecay, line 5: What is pct intended to stand for in max_payers_x_pct_distance? Could name be simplified at all?

src/movement.rs
```diff
@@ -13,10 +14,24 @@ use crate::{
     state::State, ArrivedEvent, DespawnMarker, Player, Stats,
 };
 
+// Distance after which the boundary is moved.
+//
+const LEFT_BOUNDARY_MAX_DISTANCE: f32 = 380.;
```
odecay, line 19: Consider moving to consts.rs. I don't feel strongly about it, could see case for keeping it here where its relevant for tweaks as well.

src/movement.rs
```diff
@@ -76,6 +92,16 @@ pub fn player_controller(
                 // Apply speed
                 dir = dir * stats.movement_speed * time.delta_seconds();
 
+                let new_x = transform.translation.x + dir.x;
+
+                // The dir.x condition allows some flexibility (e.g. in case of knockback), given
+                // the current state of development. To be removed once the movement logic is
+                // stabilized.
+                //
```
odecay, line 100: :+1:

src/movement.rs
```diff
@@ -50,45 +66,102 @@ pub fn player_controller(
...
-            dir.y = 0.;
-        }
+                // The dir.x condition allows some flexibility (e.g. in case of knockback), given
+                // the current state of development. To be removed once the movement logic is
+                // stabilized.
+                //
+                if dir.x < 0. && new_x < left_movement_boundary.0 {
+                    dir.x = 0.;
+                }
+                println!("{}", new_x);
```
odecay, line 104: Remember to remove in future

src/movement.rs
```diff
@@ -50,45 +66,102 @@ pub fn player_controller(
...
+                    dir.y = 0.;
+                }
+
+                //Move the player
+                Some(dir)
+            }
+        })
+        .collect::<Vec<_>>();
+
+    if player_dirs.len() == 2 {
```
odecay, line 118: We want to support 4 players, can leave for future prs.

src/movement.rs
```diff
@@ -50,45 +66,102 @@ pub fn player_controller(
...
-            state.set(State::Running);
+    for ((mut state, _, mut transform, facing_option, _), dir) in
+        query.iter_mut().zip(player_dirs.iter())
+    {
+        if let Some(dir) = dir {
+            transform.translation.x += dir.x;
+            transform.translation.y += dir.y;
+
+            //Set the player state and direction
+            if let Some(mut facing) = facing_option {
```
odecay, line 152: I must have missed it in a different merge, I don't think facing needs to be an option at all in this system. Could be changed in this pr or left to another. @zicklag I could be wrong if there is a way the default playerbundle could ever not be added.

64kramsystem:
Addressed all of them. Generalized the forward (right) limit to any number of players.

Note that the limitation on the backward (left) behavior is not relative to the rightmost player position, instead, it's based on the boundary. This can make the distance grow more than wanted. This is intended, since the limitation logic will need to be updated along with the new camera logic.

64kramsystem:
Extra note: I'm using a feature of the latest stable Rust (`f32#total_cmp`).

assets/default.game.yaml
64kramsystem: It stood for `pct`, but I've renamed it to `fraction`, which is very obvious.

src/movement.rs
64kramsystem: I've moved it, although generally speaking, I think there is value on keeping module-private constants inside the module (as they are not used anyway anywhere else).

src/movement.rs
64kramsystem: Taken care of.

src/movement.rs
64kramsystem: I've opened an issue (#63), which describes the general note, and includes a reference to this condition :smile:.

64kramsystem:
Since nobody was looking, I've added the new camera commits :grimacing: This now fully closes the first task of #29.

64kramsystem:
Re-requested review, due to new commits with the new camera logic.

src/movement.rs
odecay: > I think there is value on keeping module-private constants inside the module (as they are not used anyway anywhere else).
Yeap that makes sense to me. (If you feel like putting that one back I wont object)

src/movement.rs
zicklag: Yeah, as far as I can tell I don't see any reason the facing component wouldn't be there.

odecay: removed in https://github.com/fishfight/punchy/commit/9629e92a3eba32bb6cbfb5fe11d22c87429348da

src/movement.rs
64kramsystem: No worries, I'm not fussy :laughing:

Since there is a certain overlap between the content of `consts.rs` and the yaml config, it may be worth, in the future, thinking about where to place them.

Since camera may have customization, possibly, at some point, I'll evaluate if moving `LEFT_BOUNDARY_MAX_DISTANCE` to the yaml config.

Long story short: fine for me :laughing:

Issue #34, opened 2022-06-30
Configure Web Player Through Query String


zicklag:
As a follow up to #23 we want to allow specifying engine configuration in the query string. Configurable options should include:

- Asset URL: Allows people playing alternative games with the same web build
- Log level

PR #65, opened 2022-07-10
Use Engine Config For Menu Skip Functionality

zicklag:
I thought that an engine config parameter made sense as a spot for the new auto-start functionality added by @64kramsystem ( great idea BTW! ).

After I get #34 finished, which is my next task, this will also make it work in the web builds which could be useful.


odecay:
Agreed this makes sense.

PR #68, opened 2022-07-11
Fix Missing Label Warnings

### LINKED ISSUE #67: Warnings for unfound ordering labels in UI plugin
Seems like the labels are not getting picked up. Could be related to some incompatibility of loopless with non string labels? I feel like last time I noticed this it was something like that but not sure.

```
2022-07-11T19:20:33.902241Z  WARN bevy_ecs::schedule::graph_utils: littlefighter2::ui::handle_menu_input wants to be after unknown label: ProcessInput
2022-07-11T19:20:33.902365Z  WARN bevy_ecs::schedule::graph_utils: littlefighter2::ui::handle_menu_input wants to be before unknown label: BeginFrame
```
Any ideas @zicklag?

zicklag:
Fixes #67

According to the egui docs it seems like those labels should have existed, but debug dump of the schedule graph seems to indicate otherwise.

It seems to work just fine without the `.before()` and `.after()` flags so I think we'll just remove them and figure out what we need to do if we see issues later.

### DIFF
--- a/src/ui.rs
+++ b/src/ui.rs
@@ -1,7 +1,7 @@
 use bevy::{prelude::*, utils::HashMap, window::WindowId};
 use bevy_egui::{
     egui::{self, style::Margin},
-    EguiContext, EguiInput, EguiPlugin, EguiSettings, EguiSystem,
+    EguiContext, EguiInput, EguiPlugin, EguiSettings,
 };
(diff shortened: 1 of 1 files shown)


odecay:
Sounds good, if we do end up having issues with it we could consider opening an issue on bevy_egui or loopless.

Issue #63, opened 2022-07-10
Unify (if possible) the player translation (and limits)


64kramsystem:
Currently, there is separate logic to handle the player translation, for example, on user input vs. attacks.

Such logic should be merged; otherwise, additional translation processing (currently, limits) would need to be applied for each point where translations are manipulated.

A standard pattern for this is all movements should send an event, which is then processed, and applied (according to the rules).

When this is implemented, remove the `dir.x < 0.` condition from the `player_controller` system.

I see this as depending on #99, so it'll be clear which systems handle translation, and how they need to be modified (it's known already which they are, but it's confusing to operate right now, since we'd need to introduce some temporal dependency).

odecay:
Moving this to v0.0.4 but if anyone disagrees feel free to move it back.

64kramsystem:
> Moving this to v0.0.4 but if anyone disagrees feel free to move it back.

It makes sense to do it for 0.0.4 (because in order to do it in a very clean way, I think it requires stages in place), however, I've thought of an easy preliminary way to do it, so that we can start fixing those pesky out-of-border movements we have now :smile: 

In case I'll complete this in time for 0.0.3, I'll just create the PR and change the milestone.

odecay:
Sounds good, the stage reorder was my motivation for moving to v0.0.4 also. :+1:

PR #69, opened 2022-07-11
Refactor fighters spawning

64kramsystem:
Extract spawning types and logic from main into player.rs and enemy.rs, and extended the respective bundles to make all the components explicit.

Right now, main.rs still has a mix of different things (e.g. player belly flop attack), so other things should be moved out.

There are still some

### DIFF
--- a/src/enemy.rs
+++ b/src/enemy.rs
@@ -1,6 +1,10 @@
 use bevy::prelude::*;
 
-use crate::animation::Facing;
+use crate::{
+    animation::Facing,
+    consts,
+    metadata::{FighterMeta, FighterSpawnMeta},
+};
 
 #[derive(Component)]
 pub struct Enemy;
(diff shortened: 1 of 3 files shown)


64kramsystem:
Should the belly flop system go into `attack.rs` (hypothetically, and in a separate PR)? I think so, although the semantics of attack are a bit fuzzy (probably, other systems will be added in the future, so I think that the belly flop system should go there).

odecay:
> Should the belly flop system go into `attack.rs` (hypothetically, and in a separate PR)? I think so, although the semantics of attack are a bit fuzzy (probably, other systems will be added in the future, so I think that the belly flop system should go there).

I like this, I was starting to do something like this myself while I was considering #63. I think the attack portion and movement portion of the flop can be different systems, triggering by input should happen in attack. I'll try to make a pr soon and ask what you think.

64kramsystem:
> I like this, I was starting to do something like this myself while I was considering #63. I think the attack portion and movement portion of the flop can be different systems, triggering by input should happen in attack. I'll try to make a pr soon and ask what you think.

Splitting them is a good idea - actually, I think it will be necessary, if we unify the player translations as described [in my recent issue](https://github.com/fishfight/punchy/issues/63).

The final shape of the split depends on if/how we decide to merge the translations. If yes, I think we should also start to think in stages.

My rough idea of splitting would be:

- stage N: input is read, and it's translated to an intention-to-move message, with the current location and the motion vector
- stage N+1: move is examined, optionally processed, and optinally applied
  - for example, if the player belly flops from the edge of the screen towards outside, the processing should remove the x component of the motion vector, but leave the y one
  - stage N+2: attack collisions are tested

That's my proposal :smile:

odecay:
Yea I thinkkkk that makes sense. We will have to account for collisions from attacks in our motion vector application as well. I keep meaning to reply to this but forgetting where it was. Would you transfer above proposal to its own issue @64kramsystem?

PR #71, opened 2022-07-12
Fix Cargo Warnings

zicklag:


64kramsystem (approved):
Thanks! My toolchain did _not_ notify me of those, not sure why :thinking:

zicklag:
Just last week I heard the tip from somebody on the forum to set the "check on save: command" to `clippy` in VSCode. It's been rather helpful, because I didn't used to get clippy lints in VSCode.

PR #72, opened 2022-07-12
Add Check Recipe to Justfile

zicklag:
Shortcut to check code before pull request: `just check`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,8 @@ Before committing and opening a PR, please run the following commands and follow
 1. `cargo clippy -- -W clippy::correctness -D warnings`
 2. `cargo fmt`
 
+Or if you install [`just`](https://github.com/casey/just) you can simply run `just check`.
+
 ## MVP Spec
 
 ![861A6300-5FFD-4DDC-B4BF-0E8514F4B87C](https://user-images.githubusercontent.com/583842/161247148-0bc07089-1409-48ca-9cc8-ee1a1edddb9e.png)
--- a/justfile
+++ b/justfile
@@ -3,6 +3,10 @@
 list:
     just --list
 
+check:
+    cargo clippy -- -W clippy::correctness -D warnings
+    cargo fmt --check
+
 build:
     cargo build
 


64kramsystem:
Do you have any thoughts of adding this also to the CI? I've just added it to my RA configuration (thanks :wink:), however, adding it to the CI would prevent anybody, especially newcomers, to commit suboptimal/broken code. I think using the `correctness` suite of lints only, as in this PR, is a reasonable choice.

64kramsystem (approved):

zicklag:
Yeah, adding it to CI seems like a good idea. Running clippy can probably be a lot faster than compiling the crate, which could be good as far as CI hours goes.

64kramsystem:
> Yeah, adding it to CI seems like a good idea. Running clippy can probably be a lot faster than compiling the crate, which could be good as far as CI hours goes.
> 
> You were mentioning earlier I think that we need to be careful what we put in CI to avoid running out of minutes, do you think just adding the check to CI would be likely to get by without running into problems?
> 
> Also, could we copy how Bevy does it maybe. They've got tons of PR, etc. I'd imagine they've got to have been pretty smart about how they manged CI times.

That's an interesting topic. Bevy could surely be looked at. I think we can't escape recompiling on the CI in the future, as some - at least very basic - form of testing/checking will likely be added.

We could be clever with caches and see if they do gives us a speed. Something generally not mentioned about Github actions caching is that it's not shared across PRs. This means that new PRs are expensive in terms of CI minutes expended. There is something that can be taken advantage of (which I did at my company) - the main branch cache _is_ shared with the PRs.

In short, what we can do in case CI times grow due to build of various types, is to cache the build artefacts for each master change; the new artefacts will be shared with PRs, whose build will be shortened considerably.

Something also that can be used to save build time is `mold` (assuming Linux builds), which is The Fastest Linker™.

64kramsystem:
To summarize: I'll give a brief check of what's the resource expenditure if we add clippy. If it takes significant time, I'll play a bit with caches.

PR #73, opened 2022-07-12
Add Rust toolchain file

64kramsystem:
This doesn't affect devs who use stable by default, but helps those who use other toolchains (ie. nightly).

There is a nice side-effect in making it clear that this project expects stable, and also prevents devs from inadvertly work on nightly.


zicklag:
Using nightly compilers has actually been recommended for better compile times before, so I wonder if there is any reason to avoid working with nightly? Nightly is often my default.

I don't think that nightly rust allows you to do anything that you can't do on stable unless you explicitly enable unstable features in the source code. We could even block that with `#![forbid(unstable_features)]` maybe.

The only think I could think of that might cause some discrepancy would maybe be that the warnings you get could vary between compiler versions and clippy messages might be different.

64kramsystem:
Oh ok, I'm fine with anything. From the discussion on the channel I thought that this project was intended to be developed on stable only, but I misunderstood. I think blocking unstable features would be a good idea, just to have somewhere, in the project, that makes it clear and enforces, what's the intended target.

zicklag:
Well I'm not on the discussion channel so there could still be something I missed. :)

64kramsystem:
@zicklag just out of curiosity, do you experience the bug I describe [here](https://github.com/rust-lang/rust-analyzer/issues/12738) on your setup (nightly)?

zicklag:
Oh, actually, I've like never had rust analyzer properly infer the types on query iterations. I didn't realize it was related to nightly. I thought I'd tried it on stable before and it didn't work either so I thought it was just a temporary limitation.

I'll have to try on stable. I've even ended up annotating the types in the code before just to make RA work so that it was easier to work with.

zicklag:
Look at that, it works on stable! Looks like I'll be using stable for now, then. :) Thanks for the tip!

lecoqjacob:
I know this issue is closed, but how did yall solve it? I switched to stable rust, and cleared my system of RA and then reinstalled via VSCode. It still doesn't infer types for bevy. What is weird is it infers types for internal things like `Entity`, but not my components.
CleanShot 2022-07-15 at 14 38 05@2x
CleanShot 2022-07-15 at 14 38 11@2x
CleanShot 2022-07-15 at 14 38 14@2x

64kramsystem:
> I know this issue is closed, but how did yall solve it?

On my setup, I just had to move to stable :grimacing: I suggest to open an issue on the RA issue tracker - they may at least give you some pointers.

zicklag:
Yeah, same here. All I did was `rustup override set stable` in the workspace, `cargo clean` and then restart vscode.

PR #75, opened 2022-07-12
Rename Crate to `punchy`

zicklag:
Good to make it consistent with the repo I think. This doesn't change much, but it does make setting the log level for the punchy module much more intuitive.


odecay:
Nice, that reminds me our window title/appsettings were not getting applied correctly, fixing that now as well.

PR #76, opened 2022-07-13
Add Clippy correctness checks to CI (WIP)

64kramsystem:
NOTE: currently testing caching on a separate repo; this PR works, but has no caching.

Hm. It takes, as of now, approximately 6 minutes to compile (and check, which takes a negligible time).

This can likely be made very fast by caching the target and cargo dir on master pushes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,3 +13,14 @@ jobs:
         with:
           command: fmt
           args: --all -- --check
+  clippy_correctness_checks:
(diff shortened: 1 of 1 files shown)


64kramsystem:
> One thing I was thinking is that we should probably put `#![warn(clippy::correctness)]` at the top of `main.rs` so that people will get warnings in their workspace about the lints, and not just when the run `just check` or get blocked by CI.

Yes, makes total sense :+1:.

PR #78, opened 2022-07-13
CI: Add Clippy correctness checks (and Cargo check data caching)

64kramsystem:
Add Clippy correctness checks. In order to make CI run in reasonable time (tens of seconds), a cache workflow has been added, which runs on each master push.

Some notes:

- the master workflow is required, because Github Actions cache is _not_ shared between PRs, but only (within a single PR, or between master and the PRs)
- the other reason to add the master caching workflow is that it can be trivially extended to cache also the build, in case we'll add testing or similar, in the future

This CI will build slowly; after merging, subsequent PRs will run fast.

For example runs/timings, see:

- master workflow run: https://github.com/64kramsystem/punchy-dev/actions/runs/2664626729
- PR workflow run: https://github.com/64kramsystem/punchy-dev/actions/runs/2664677214

### DIFF
--- a/.github/workflows/build-caches.yml
+++ b/.github/workflows/build-caches.yml
@@ -1,0 +1,28 @@
+name: Build caches
+
+on:
+  push:
+    branches:
+      - master
+jobs:
+  # Watch out! Clippy (cache) data is shared with build data, but it's not the same.
+  # In order to cache build, add a separate job (or extend this).
+  build_clippy_cache:
+    name: Build Clippy cache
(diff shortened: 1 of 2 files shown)


64kramsystem:
I've created a new PR because I've closed the last one by accident, and could not be reopened due to changes.

zicklag (approved):
Those times are beautiful, and the workflow didn't end up complicated or anything.

Great job! I'm going to have to use this trick on my repos. :wink:

PR #80, opened 2022-07-13
Add audio base support, with enemy attack effect(s)

### LINKED ISSUE #55: First audio pass
For game music we've already got a completed OST that can be reused as-is for this game:
https://spicylobster.itch.io/fish-world-pack

We'll probably follow up with a beatemup-tailored remix of this album way later.

For SFX, anything creative-commons is on the table for the purpose of temporary use. A [quick search for 'punch'](https://opengameart.org/art-search-advanced?keys=punch&title=&field_art_tags_tid_op=or&field_art_tags_tid=&name=&field_art_type_tid%5B%5D=13&field_art_licenses_tid%5B%5D=17981&field_art_licenses_tid%5B%5D=2&field_art_licenses_tid%5B%5D=17982&field_art_licenses_tid%5B%5D=3&field_art_licenses_tid%5B%5D=6&field_art_licenses_tid%5B%5D=5&field_art_licenses_tid%5B%5D=10310&field_art_licenses_tid%5B%5D=4&field_art_licenses_tid%5B%5D=8&field_art_licenses_tid%5B%5D=7&sort_by=score&sort_order=DESC&items_per_page=24&Collection=) resulted in these two which I quite like:
https://opengameart.org/content/punch-slap-n-kick
https://opengameart.org/content/37-hitspunches

Also:
https://www.kenney.nl/assets/interface-sounds
https://www.kenney.nl/assets/ui-audio

64kramsystem:
Base audio support (see #55):

- audio setup (two channels)
- fighter (player/enemies) effects, per-frame
- main menu music
- level music

The per-frame implementation was actually not trivial to design, although it fits smoothly. It's been necessary for states/attacks like the belly flop.

Closes #55.


64kramsystem:
@zicklag I can't make the effect play when I use the asset loader (the sound framework as I've implemented works; it's a problem with asset loading). Can you help me figure out what's wrong?

If I load the asset directly (see HEAD commit), the sounds play, but without that (just comment [this](https://github.com/64kramsystem/punchy-dev/blob/add_sound/src/attack.rs#L141) line), the effect won't play. The asset seems to be loaded though, which is what confuses me.

zicklag:
Sure, give me a second and I'll look at it.

zicklag:
You just have to add the sound asset paths to the dependencies for the fighter asset. I pushed a commit that does it.

64kramsystem:
This should be enough to for the first pass. I used random assets, since they're fully configurable.

src/assets.rs
```diff
@@ -111,7 +119,8 @@ impl AssetLoader for GameMetaLoader {
                 LoadedAsset::new(meta)
                     .with_dependencies(vec![start_level_path, main_menu_background_path])
                     .with_dependencies(locale_paths)
-                    .with_dependencies(font_paths),
+                    .with_dependencies(font_paths)
+                    .with_dependencies(vec![music_path]),
```
zicklag, line 123: We could move `music_path` into the vec created a couple lines above it so that we don't allocate another vector.

src/attack.rs
```diff
@@ -95,17 +97,23 @@ fn player_attack(
...
     mut event_reader: EventReader<ArrivedEvent>,
     mut commands: Commands,
+    fighter_assets: Res<Assets<FighterMeta>>,
 ) {
     for event in event_reader.iter() {
-        if let Ok(mut state) = query.get_mut(event.0) {
+        if let Ok((entity, mut state, fighter_handle)) = query.get_mut(event.0) {
+            let fighter = fighter_assets
+                .get(fighter_handle)
+                .expect("Fighter asset not found");
```
zicklag, line 109: I think it is possible that the fighter asset isn't loaded at this point, though I'll admit it's unlikely that the user manages to press the attack button before the figher asset finishes loading.

Still, I think we should probably put this in an `if let Some(fighter) = fighter_assets.get(fighter_handle {}`, or if it's more readable, to avoid indenting everything we could do this:

```rust
let fighter = if let Some(fighter) = fighter_assets.get(fighter_handle) {
    fighter
} else {
    // Skip this iteration of the loop
    continue;
};
```

src/audio.rs
```diff
@@ -0,0 +1,106 @@
...
+        // The safest way to remove the sound component is on the next state, because the component
+        // can be remove only at the last frame of animation, which in theory, may be skipped if
+        // there is an unexpected lag.
+        // Alternatively, we could just not care, since subsequent states+effects will overwrite
+        // the component.
+        if *fighter_state != state_effects.state {
+            commands
+                .entity(entity)
+                .remove::<FighterStateEffectsPlayback>();
+            return;
```
zicklag, line 61: If I understand correctly, this should be a `continue`, not a `return`, so that the remaining fighters are still looped through.

src/main.rs
```diff
@@ -492,26 +504,41 @@ fn unpause(mut commands: Commands, input: Query<&ActionState<MenuAction>>) {
...
         if *state != State::Attacking {
             if *state != State::Idle && *state != State::Running {
                 return;
             }
             if input.just_pressed(PlayerAction::FlopAttack) {
                 state.set(State::Attacking);
+
+                let fighter = fighter_assets
+                    .get(fighter_meta)
+                    .expect("Fighter asset not found");
```
zicklag, line 534: Let's use an `if let` instead of panicking here.

zicklag (commented):
Looking good!

I left a couple comments, and I pushed a commit to fix an issue where the menu music wasn't playing if you started the game and came back to the menu.

I also figured out why the menu music play system was running twice. It was because the menu play system was added once as an enter system, and another time as a part of the main menu condition set.

64kramsystem:
> Also, just let me know if you don't like me pushing commits to your branch. :) I don't want to feel pushy. ( Pun not initially intended upside_down_face )

Oh don't worry, I actually find it very convenient.

I'll review the changes later. I've spent hours on ggez issues today :man_facepalming:.

src/audio.rs
64kramsystem: Good catch :smile:. Fixed.

zicklag (approved):

64kramsystem:
Changed the two music tracks according to the channel discussion; merging.

Issue #83, opened 2022-07-14
Bottle throwable item


zicklag:
### WIP Spec ( Needs Review )

The bottle item should:

- [x] Be spawned on the map, laying on the floor in places determined by the level info.
- [x] Be able to be grabbed by a fish
- [x] Be able to be thrown after being grabbed
- [x] Should break after thrown

This would also change the player moves:

- [x] Player can no longer throw an unlimited amount of bottles
- [x] Player has a move that is pick up and throw using the same button

zicklag:
@erlend-sh @odecay I wrote a work-in-progress spec for the bottle item. I had to make some guesses as to how we wanted to handle it. Does that look right to you?

Also, when you throw the item, does it shoot straight forward, or thown in an arch? I feel like the bottle should throw in an arch.

erlend-sh:
Sounds right 👍 
Throws should be slightly arched, yep.

64kramsystem:
Point 4 (Should break after thrown) may not be trivial.

With the current design, within one frame/stage, systems are not aware of each other. In short, the shoot system can't know that the throw system executed, so if a user taps both shoot and throw keys in a single frame, a bottle will be thrown and shot.

If it's not possible to trigger two action keys at the same time, then both systems will not run in the same frame, but this leaves an very bug-prone logic.

This problem will be solved once there is a single entry point for testing the keypresses and deciding which action to take, rather than having systems individually testing the keypress.

64kramsystem:
> Player has a move that is pick up and throw using the same button

Since shoot works on bottle as well, I've made the PR use also the shoot button to pick up items. This can be trivially changed, though.

odecay:
> > Player has a move that is pick up and throw using the same button
> 
> Since shoot works on bottle as well, I've made the PR use also the shoot button to pick up items. This can be trivially changed, though.

I think ultimately we should not allow bottles to be "shot", that sort of projectile attack will be reserved for bows or guns or other such projectile weapons. All weapons should additionally be able to be thrown, melee weapons included.

This has some implications for the inputs we use, I feel like it makes sense for melee weapons to default to picking them up either with a dedicated pickup button or with the attack button, but I think it could cause issues when "mashing" that button if it also causes a throw to happen as soon as the item is picked up, which may only be desired behavior for dedicated throwing weapons (even then unsure if this is beneficial). I advocate for pickup and attack to be on the same button, and throw to be on a 2nd button.

[...]

PR #100, opened 2022-07-16
Input Settings Menu

zicklag:
It's working! 

Opening as a draft because I need to do a self-review, but it's functional. 🎉

setings-menu-3

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ getrandom = { version = "0.2", features = ["js"] }
 bevy_mod_debugdump = { version = "0.4", optional = true }
 bevy-inspector-egui = { version = "0.11", optional = true }
 bevy-inspector-egui-rapier = { version = "0.4", optional = true, features = ["rapier2d"] }
-leafwing_input_manager = { git = "https://github.com/Leafwing-Studios/leafwing-input-manager.git", rev = "e8bd6e0" }
+leafwing_input_manager = { git = "https://github.com/zicklag/leafwing-input-manager.git", branch = "backport-leafwing-dev-to-bevy-0.7" }
 unic-langid = "0.9.0"
 bevy_fluent = { git = "https://github.com/kgv/bevy_fluent", rev = "d41f514" }
 sys-locale = "0.2.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,6 +9,7 @@ anyhow = "1.0.58"
 bevy = "0.7.0"
 bevy-parallax = "0.1.2"
 bevy_egui = "0.14"
+egui_extras = "0.18.0"
 bevy_kira_audio = { version = "0.10.0", features = ["mp3"] }
 bevy_rapier2d = { version = "0.14.1", features = ["debug-render"] }
 iyes_loopless = "0.6.1"
--- a/src/ui/main_menu.rs
+++ b/src/ui/main_menu.rs
@@ -14,7 +14,7 @@ use crate::{
 
 use super::{
     widgets::{bordered_button::BorderedButton, bordered_frame::BorderedFrame, EguiUIExt},
-    EguiUiExt,
+    EguiContextExt, EguiResponseExt, WidgetAdjacencies,
 };
 
 #[derive(Component)]
@@ -94,6 +94,7 @@ pub fn main_menu_system(
     menu_input: Query<&ActionState<MenuAction>>,
     mut app_exit: EventWriter<AppExit>,
     mut storage: ResMut<Storage>,
+    mut adjacencies: ResMut<WidgetAdjacencies>,
 ) {
     let menu_input = menu_input.single();
 
@@ -147,6 +148,7 @@ pub fn main_menu_system(
                             &mut settings,
                             &mut menu_state,
                             &mut storage,
+                            &mut adjacencies,
                             tab,
                             &localization,
                             &game,
@@ -183,13 +185,8 @@ fn main_menu_ui(
             &localization.get("start-game"),
         )
         .min_size(min_button_size)
-        .show(ui);
-
-        // Focus the start button if nothing else is focused. That way you can
-        // play the game just by pressing Enter.
-        if ui.memory().focus().is_none() {
-            start_button.request_focus();
-        }
+        .show(ui)
+        .fous_by_default(ui);
 
         if start_button.clicked() || engine_config.auto_start {
             commands.insert_resource(game.start_level_handle.clone());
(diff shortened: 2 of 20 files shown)


Cargo.toml
```diff
@@ -22,7 +23,7 @@ getrandom = { version = "0.2", features = ["js"] }
 bevy_mod_debugdump = { version = "0.4", optional = true }
 bevy-inspector-egui = { version = "0.11", optional = true }
 bevy-inspector-egui-rapier = { version = "0.4", optional = true, features = ["rapier2d"] }
-leafwing_input_manager = { git = "https://github.com/Leafwing-Studios/leafwing-input-manager.git", rev = "e8bd6e0" }
+leafwing_input_manager = { git = "https://github.com/zicklag/leafwing-input-manager.git", branch = "backport-leafwing-dev-to-bevy-0.7" }
```
zicklag, line 26: Note: Because I had to make a fix to leafwing, and they've already migrated to the pending bevy 0.5, there's no way around using my temporary fork for this anymore. I think we'll just have to wait 'till we migrate to Bevy 0.8 to switch this to a more permanent release.

zicklag:
Just did a more thorough review of my own code: added lots of comments, removed some unneeded code paths, and fixed a minor usability issue with button hovering and focusing. I think this is ready now.

odecay:
Gonna try to review this tomorrow. 

One thing to consider for a future pr is a "bind all" mode which when entered prompts for input to bind for the first Action on list, then immediately progresses to the next Action after pressing a key/button until all actions are bound to buttons.

64kramsystem:
Going to give a review as well tomorrow or the day after, as I'm about to travel :smile:

src/ui/main_menu.rs
```diff
@@ -0,0 +1,768 @@
...
+    }
+}
+
+/// Render the sound settings UI
+fn sound_settings_ui(_ui: &mut egui::Ui, _game: &GameMeta) {
+    // This is un-reachable right now
+    todo!("Implement sound settings UI");
+}
+
+/// Format an InpubKind as a user-facing string
```
odecay, line 676: nbd but InputKind

odecay (commented):
Looks good! I didn't have much meaningful feedback, noticed few small things that looked like spelling errors.

None of this should block this work or be included in this pr but some thoughts came up while looking it over.

I think maybe we will want to think more about what options players have for binding in the menu, I feel like maybe it should allow for selecting an input type and then only showing a map for that type instead of gamepad + keyboard 1/2 at the same time. 

This might also relate to how we expect bindings to work in multiplayer both in local multiplayer and networked multiplayer. Which player gets to set their bindings once the menu is triggered? Right now it sets for all players correct?

zicklag:
> noticed few small things that looked like spelling errors.

Good catch, I fixed those now. :+1: 

> I feel like maybe it should allow for selecting an input type and then only showing a map for that type instead of gamepad + keyboard 1/2 at the same time.

That's a great idea. We could definitely do that in a next iteration. It would be an easy update.

> This might also relate to how we expect bindings to work in multiplayer both in local multiplayer and networked multiplayer. Which player gets to set their bindings once the menu is triggered? Right now it sets for all players correct? 

Yep. Everybody player shares the same gamepad bindings, and there are the two keyboard bindings so that you can play two players with one keyboard.

If we added a way to rotate through which bindings you are actually setting, like you mentioned, then that would make it easier to set different gamepad bindings for player 1 and player 2.

I'd have done separate gamepad bindings for both gamepads if it weren't for the fact that I ran out of room. :) Selecting which binding you are setting first solves a lot of problems.

odecay:
> > noticed few small things that looked like spelling errors.
> 
> Good catch, I fixed those now. +1
> 
> > I feel like maybe it should allow for selecting an input type and then only showing a map for that type instead of gamepad + keyboard 1/2 at the same time.
> 
> That's a great idea. We could definitely do that in a next iteration. It would be an easy update.
> 
> > This might also relate to how we expect bindings to work in multiplayer both in local multiplayer and networked multiplayer. Which player gets to set their bindings once the menu is triggered? Right now it sets for all players correct?
> 
> Yep. Every player shares the same gamepad bindings, and there are the two keyboard bindings so that you can play two players with one keyboard.
> 
> If we added a way to rotate through which bindings you are actually setting, like you mentioned, then that would make it easier to set different gamepad bindings for player 1 and player 2.
> 
> I'd have done separate gamepad bindings for both gamepads if it weren't for the fact that I ran out of room. :) Selecting which binding you are setting first solves a lot of problems.

Yep all totally makes sense. Lets add an issue for next steps for input binding after we get this PR merged in. Feel free to merge whenever you feel its ready.

zicklag:
> Feel free to merge whenever you feel its ready.

Since nothing I'm doing next needs this I'll wait a day to let @64kramsystem look at it.

zicklag:
> Would you mind if I took that one?

Oh, not at all, go ahead. That one looked easy and I was trying to find what the most important things to get out before the next release were, but I'd rather you took it anyway.

> What are you planning on working on next btw?

Right now I'm working on a quick experiment to see if an idea I had regarding scripting is feasible or not.

I figured that if we have a solution for scripting, it might have an impact on how we program the players or items, so it'd be good to have some idea of what that might look like earlier rather than later.

It should just take a day or two to see if my idea has any merit, and it's kind of fun experimentation work so I figured I'd take a quick look as sort of a "break".

Then I'll come back to the #83 probably next.

odecay:
> > Would you mind if I took that one?
> 
> Oh, not at all, go ahead. That one looked easy and I was trying to find what the most important things to get out before the next release were, but I'd rather you took it anyway.
> 
> > What are you planning on working on next btw?
> 
> Right now I'm working on a quick experiment to see if an idea I had regarding scripting is feasible or not.
> 
> I figured that if we have a solution for scripting, it might have an impact on how we program the players or items, so it'd be good to have some idea of what that might look like earlier rather than later.
> 
> It should just take a day or two to see if my idea has any merit, and it's kind of fun experimentation work so I figured I'd take a quick look as sort of a "break".
> 
> Then I'll come back to the #83 if nobody else wants it, or I'll just work on the update binding menu.

okay perfect! scripting sounds awesome, followed along with the threads/issues on mun scripting/bevy_mod_scripting, were you thinking something like #107?

zicklag:
Replied here to make it more visible: https://github.com/fishfight/punchy/issues/107#issuecomment-1189447103

64kramsystem:
> > Feel free to merge whenever you feel its ready.
> 
> Since nothing I'm doing next needs this I'll wait a day to let @64kramsystem look at it.

Looks good to me (although I'm not very familiar with the areas :grimacing:) :smile:

64kramsystem (approved):
:+1:

PR #101, opened 2022-07-16
Add PlayerIndex component, and fix HUD players ordering

### LINKED ISSUE #81: Add Player Indexes
Players should have a component such as `PlayerIndex(usize)` on them that just sets which player they are.

This can be used where the player order is important, for example, in the HUD display, which happens to change the ordering when a fish get's punched sometimes, just due to the unstable iteration order over player entities.

64kramsystem:
Add the PlayerIndex component, and fixes the HUD list of players not being stable.

A potential refactoring is to draw the player images in the HUD, using a function of their index, rather than iterating them and adding spaces for each iteration.

Closes #81.

### DIFF
--- a/src/player.rs
+++ b/src/player.rs
@@ -11,9 +11,13 @@ use crate::{
 #[derive(Component)]
 pub struct Player;
 
+#[derive(Component)]
+pub struct PlayerIndex(pub usize);
+
 #[derive(Bundle)]
 pub struct PlayerBundle {
     player: Player,
+    index: PlayerIndex,
     facing: Facing,
     #[bundle]
     transform_bundle: TransformBundle,
@@ -43,6 +47,7 @@ impl PlayerBundle {
 
         PlayerBundle {
             player: Player,
+            index: PlayerIndex(player_i),
             facing: Facing::Right,
             transform_bundle,
             fighter_handle,
--- a/src/ui/hud.rs
+++ b/src/ui/hud.rs
@@ -30,19 +30,21 @@ pub fn render_hud(
     let mut players = players.iter().collect::<Vec<_>>();
     players.sort_by_key(|(player_i, _, _)| player_i.0);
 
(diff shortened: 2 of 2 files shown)


zicklag:
What do you think about putting the index in the `Player` component?

I'm not sure if there's a reason to do one or the other, but it did occur to me that we might not need a `PlayerIndex` component and a `Player` component.

64kramsystem:
@zicklag do you prefer more functional approaches like [this](https://github.com/64kramsystem/punchy-dev/commit/9327444fdb7ced522f3bb3ab7ba94feceef2a749) (I do)?

zicklag:
> @zicklag do you prefer more functional approaches like [this](https://github.com/64kramsystem/punchy-dev/commit/9327444fdb7ced522f3bb3ab7ba94feceef2a749) (I do)?

Yeah, that's good. :+1: 

I don't always do it, but I like the functional style.

64kramsystem:
> What do you think about putting the index in the `Player` component?
> 
> I'm not sure if there's a reason to do one or the other, but it did occur to me that we might not need a `PlayerIndex` component and a `Player` component.

Hmmm, that's an interesting point. I guess that that approach trades off strict semantics (I think `Player` as the typical empty marker component, rather that a stateful one) for simplicity (one component less :grimacing:).

The implication of moving the index into the `Player` component is that index is a special attribute, as opposed to, for example, score/name/etc. :thinking: 

Interesting question, I'm not sure of either. It depends on how we imagine the future design. We could move the index into the Player, then if we'll add a new attribute which we don't deem "less special" than index, we move the index into a component.

zicklag:
I'm good with either one really. I'd say it's good in a separate component for now. It's not like it's hard to change later no matter how we do it now.

64kramsystem:
Ok! If it passes your review, I'll merge it :smile:

zicklag (approved):

PR #104, opened 2022-07-16
Remove orphan Panning component and related code

### LINKED ISSUE #61: Camera panning and player move both happens if panning key combination includes the player move key
If a camera panning key combination (e.g. Ctrl+Up) includes a key from player move (e.g. Up), both actions will be performed.

64kramsystem:
Remove the now orphan Panning component and related code.

Also remove the controls section from the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,19 +14,6 @@ https://user-images.githubusercontent.com/1379590/175423671-0dc7bea3-6c63-42ed-a
 
 https://fishfight.github.io/punchy/demo/
 
-## Controls
-
-| Action            | ⌨️ Keyboard                      | ⌨️ Player 2               | 🎮 Gamepad       |
-| ----------------- | ------------------------------- | ------------------------ | ---------------- |
-| Movement          | W / S / A / D                   | Up / Down / Left / Right | Left Stick, DPad |
-| Throw Item        | C                               | Period                   | West             |
-| Shoot             | V                               | Right Shift              | East             |
-| Flop Attack       | Spacebar                        | Comma                    | South            |
-| Zoom Camera       | Ctrl + + / -                    | ❌️                       | ❌️               |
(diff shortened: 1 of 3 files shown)


zicklag:
I touched a lot of the input stuff with #100 including removing the camera controls so it's probably easier to do this after that merges so that I don't have to deal with merge conflicts on the big PR.

I think I've almost got that finished, just needs a couple small fixes to the button adjacency map.

I removed most of the unneeded camera stuff in #100 but I see you also modified `camera_follow_player` so that might still need to be done.

src/camera.rs
```diff
@@ -80,14 +17,14 @@ pub fn camera_follow_player(
...
-        let (mut camera, panning) = camera_query.single_mut();
+        let camera = camera_query.single();
 
         let max_player_x_diff =
             max_player_x - camera.translation.x - game_meta.camera_move_right_boundary;
 
         if max_player_x_diff > 0. {
             // The x axis is handled by the parallax plugin.
-            camera.translation.y = consts::GROUND_Y + panning.offset.y;
+            // The y axis value doesn't change.
```
zicklag, line 27: These changes should still be made, but most of the other changes were already done in #100.

README.md
```diff
@@ -20,8 +20,6 @@ https://fishfight.github.io/punchy/demo/
 | Throw Item        | C                               | Period                   | West             |
```
zicklag, line 20: Maybe we should totally remove the controller input binding list here?

Users can now view/change the bindings in the game, and it's always easy to forget to update the readme when we add/remove a new move or we change the bindings.

zicklag (changes requested):
Now that #100 was merged, most of this is already done, but there are couple things in this PR that we still want to do.

Though it might be easier to open a totally new PR.

README.md
odecay: I think it would be fine to remove now that its possible to see in the controller settings menu.

zicklag (approved):

Issue #102, opened 2022-07-16
Project is slow to compile; double check that there isn't anything unintentionally slow 


64kramsystem:
### Description

Compiling the project is (relatively) very time consuming. Even making small changes causes slow compiles on my (8 threads) laptop - 35+ seconds. I'm even using the mold linker, which is as fast as it gets.

I have no idea what causes such slowdown. Compiling could be profiled, but maybe somebody has an idea (based on experience) of why it takes so long.

### Alternatives & Prior Art

I've tried the `dynamic` Bevy feature, but it doesn't bring any improvement.

zicklag:
I've noticed the same thing. My times are much better with 12 cores ending up 14 seconds at worst, and often better depending on the change, but it's not at the usual bevy project of around roughly 2-6 seconds.

The only think I can think of right off the bat is that I always put all my code in `lib.rs` ( and submodules ) and the `main.rs` file is always a shim that simply calls `my_crate::run()`.

I do it because you can't integration test `main.rs` because it's outside the crate, but maybe it also effects compile times.

It think the times are better on nightly with the shared-generics flag.

I've also tried LLD and mold in an attempt to improve times, but apparently the blocker isn't the linker. I think there's something like a `-Z timings` flag you can pass to Cargo to get some stats, but I'll have to see exactly how to do it because I can't remember.

alice-i-cecile:
Some very good compilation time debugging advice [here](https://fasterthanli.me/articles/why-is-my-rust-build-so-slow) by `fasterthanlime` and in the [followup article](https://fasterthanli.me/articles/when-rustc-explodes).

Please let us know what helped!

alice-i-cecile:
`lto = true`: this will cause extremely slow release builds. Are you seeing this problem for debug builds too?

64kramsystem:
> `lto = true`: this will cause extremely slow release builds. Are you seeing this problem for debug builds too?
> 
> In general, I would only turn on `lto` for final releases that are going to be in the hands of players.

Hello! Yes, I'm experiencing this on debug (default) builds.

64kramsystem:
> Some very good compilation time debugging advice [here](https://fasterthanli.me/articles/why-is-my-rust-build-so-slow) by `fasterthanlime` and in the [followup article](https://fasterthanli.me/articles/when-rustc-explodes).
> 
> Please let us know what helped!

[...]

PR #109, opened 2022-07-17
Disable LTO entirely for dev profile

64kramsystem:
Even thin LTO (dev default) takes around 30/40% of the compilation time. Since there is not performance pressure right now (in dev), it can be entirely disabled.

See #102.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,6 +45,7 @@ debug = false
 
 [profile.dev]
 opt-level = 1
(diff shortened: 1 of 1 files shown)


zicklag (approved):
Sweet!

64kramsystem:
This may require clearing the Github action cache. I'll observe the effect on the future PRs, then take action if required.

PR #111, opened 2022-07-20
Build for different platforms during CI

orhun:
This PR extends the continuous integration workflow to add a build check for different platforms. It is for making sure that the game builds fine for each push to `master`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,6 @@ jobs:
           - { os: macos-latest, target: 'x86_64-apple-darwin' }
           - { os: macos-latest, target: 'aarch64-apple-darwin' }
           - { os: windows-latest, target: 'x86_64-pc-windows-msvc' }
-          - { os: windows-latest, target: 'i686-pc-windows-msvc' }
     steps:
       - name: Checkout
         uses: actions/checkout@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,9 @@
 name: CI
 
 on:
+  push:
+    branches:
+      - master
   pull_request:
 
 jobs:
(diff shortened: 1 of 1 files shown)


64kramsystem:
Will check this tomorrow. Sorry for being vague about the timeline, but yesterday I've travelled, so today it was very messy :grimacing:

64kramsystem (commented):
Which are your thoughts on executing this on PRs (and making the jobs success a requirement) rather than master? This would block issues before they reached master, rather than after landing in master.

But overall, good for me. We'll deal with time expenditure if it will be a problem.

zicklag (approved):
This looks good to me. @64kramsystem it looks like this _does_ run on pull requests, so I think that satisfies your above comment.

It also additionally runs on master, which may actually have different, untested commits because of merge commits. For instance, if we have two pull requests and CI works for both of them, and we merge both of them one after the other, we don't technically know that CI would pass for the combination of the two.

That is something we may actually care about once we get a little more serious about our workflow and we have the game more established.

Two possible solutions to the problem are:

1. Only allow fast-forward commits to master. This is a simple and easy solution that comes with the downside of requiring us to have all pull requests merged serially, which can cost you some time manually waiting for CI and merging one PR after the other, if you have multiple PRs to get done.
2. The other common option is to use a bot like bors to do the merging and batching for you. If you have five pull requests and you tell bors to merge them all, it will combine them into one batch, run CI on it to make sure it passes, and then merge to master. If they fail combined, it will use it's algorithm to take out commits until it finds the one that causes the failure, trying to use as few builds as possible to get to a solution.

I've setup Bors on my repos before and it's not hard to use or get setup, so it's probably the best option for us moving forward. It's just a matter of when we decide we want to have some more stability as far as CI and master merges goes. Probably want that as soon as we get more common contributors.

zicklag:
I'm going to merge this now, but feel free to follow up with changes if you think any are necessary.

64kramsystem:
> @64kramsystem it looks like this _does_ run on pull requests

Right! I missed the code above :man_facepalming:.

Issue #103, opened 2022-07-16
Ready for Fish Launcher


erlend-sh:
We’d like to initialize Punchy through https://github.com/fishfight/FishLauncher

We use a convention for release assets and they should follow this to work seamlessly with the launcher. Explained here: https://github.com/fishfight/FishFight/blob/main/PACKAGING.md#binary-releases

@orhun can answer any further questions.

zicklag:
This should be trivial. Bevy games follow that exact convention by default.

I haven't looked at the launcher itself yet, but it should be nice and easy to integrate.

zicklag:
After looking a little bit more into it, it seems like all we really need to do is setup a release pipeline on GitHub actions, very similar to this one from FishFight: https://github.com/fishfight/FishFight/blob/main/.github/workflows/release.yml.

Nuances such as dependencies, file names, and some similar things will be different, but it should be a straight-forward migration of that script.

orhun:
Yup, eggxactly 🥚 I will take a look at the repository and adopt a similar workflow for release builds. It should be fairly straightforward.

zicklag:
I'm closing this because I think that we've done everything we need to do on the Punchy side of this. Feel free to re-open if there's other stuff we need to do!

PR #112, opened 2022-07-20
Add GitHub workflow for automated binary releases

orhun:
This PR adds a release workflow for creating binary archives when a tag is pushed. It is required for the launcher, see #103

As an example, see the release assets created on my fork: https://github.com/orhun/punchy/releases/tag/v0.0.3

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -1,0 +1,125 @@
+name: CD
+
+on:
+  push:
+    tags:
+      - 'v*'
+
+jobs:
+  build_release:
+    name: Build release binaries
+    runs-on: ${{ matrix.config.os }}
+    continue-on-error: true
+    strategy:
+      fail-fast: false
+      matrix:
+        config:
+          - { os: ubuntu-latest, target: 'x86_64-unknown-linux-gnu' }
+          - { os: macos-latest, target: 'x86_64-apple-darwin' }
+          - { os: macos-latest, target: 'aarch64-apple-darwin' }
+          - { os: windows-latest, target: 'x86_64-pc-windows-msvc' }
+          - { os: windows-latest, target: 'i686-pc-windows-msvc' }
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Set the release version
+        shell: bash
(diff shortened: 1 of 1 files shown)


zicklag (approved):
Awesome, thanks!

odecay:
I'm wondering if it may be worthwhile axing the windows i686 build to save CI time?

erlend-sh:
I’m on board with axing windows i686. They have the option of building themselves or using the web build anyhow.

64kramsystem (approved):
Ok, this can go in also for me, with the i686 removed.

-----

We'll see how much resources this takes. If it takes too much resources, we have a couple of things that can help. For the "future selves", in case we'll need them, they are:

1. disable caching, since the caches will likely not be effective anyway; this way, we'll avoid them interfere with the `cargo check` caches
2. use the `master` branch cache for the ubuntu build.

64kramsystem (approved):
Good to merge for me :smile: :+1:

orhun:
PR is updated in 8df4080929dc029d59d678447e04009822b6cae7

This means that users who run the launcher on Windows i686 won't be able to play `punchy`, but that's rather a minority so I agree with removing i686 for improving the build times.

64kramsystem:
Well, turns out, I was wrong on the cache size :star_struck: 

I've confused Github "storage" with "caches", which are separated. Cache limit is [10 GB](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy), so we don't need to worry at all :sweat_smile:

64kramsystem:
Ok, it seems that I was also wrong on the available workflow running time available 🤬

It seems that public repositories don't have time budgets, but only constraints on how many runners run concurrently (see [here](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits)).

In short: we won't have any problem unless we the PR workflows start to pile up (which I think won't happen, at least for a long time) :smile:

zicklag:
Ah, that makes a lot of sense how Bevy get's away with it then! With 284 open pull requests I thought they'd have run out by now. :laughing:

PR #115, opened 2022-07-20
Update badges in README.md

orhun:


64kramsystem (approved):
Good for me, although, the source would be more readable if there was one link per line (the rendering should be unaffected, because there are no line breaks), since with a single line, in order to understand which badges are there, one needs to fully scan it.

zicklag (approved):

PR #118, opened 2022-07-21
Return to main menu on players death

### LINKED ISSUE #96: Program doesn't return to main menu when players lose
When the players lose, the program doesn't return to the main menu (currently, it crashes).

While it's probably intended right now not to spend effort on this, I think it'd be nice to just return to the main menu on game over. This may (not 100% sure) have also some architectural consequences, if it requires some cleanup to be designed/implemented (which I think is g00d).

64kramsystem:
This PR avoids getting the game getting stuck (without players), until an actual game over state is added.

Closes #96.


zicklag (approved):
looks good :+1:

PR #119, opened 2022-07-21
Move attack systems into attack plugin (module)

64kramsystem:
Minor refactoring, while I was studying the systems for the stages design.

### DIFF
--- a/src/attack.rs
+++ b/src/attack.rs
@@ -5,7 +5,7 @@ use bevy::{
     hierarchy::{BuildChildren, DespawnRecursiveExt},
     math::Vec2,
     prelude::{
-        default, App, AssetServer, Assets, Commands, Component, Entity, EventReader, Handle,
+        default, App, AssetServer, Assets, Commands, Component, Entity, EventReader, Handle, Local,
         Plugin, Query, Res, Transform, With,
     },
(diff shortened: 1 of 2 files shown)


src/attack.rs
```diff
@@ -31,7 +31,15 @@ pub struct AttackPlugin;
...
-        app.add_system(player_attack.run_in_state(GameState::InGame));
+        // Can't be currently converted to a ConditionSet, since (it seems that) systems inside
+        // don't have temporal methods available (e.g. after()).
+        app.add_system(player_attack.run_in_state(GameState::InGame))
+            .add_system(player_flop.run_in_state(GameState::InGame))
+            .add_system(
+                enemy_attack
+                    .run_in_state(GameState::InGame)
+                    .after("move_to_target"),
+            );
```
odecay, line 42: Gonna merge this because I want it in now/I am touching this code now too, but may return in future to look at this plugin/conditionset thing and see if its possible to set it up with loopless

odecay (approved):

src/attack.rs
64kramsystem: Sure! If/once we split in stages, the `after` relationship won't be needed anymore, since it will be implicit in the stages ordering.

PR #120, opened 2022-07-22
Spawn items support

64kramsystem:
Add spawn items support:

- metadata types, and metadata loading
- (level) loading and instantiation

The level loading (`main.rs#load_items`) follows the pattern of loading items metadata, then separately instantiating them.

I've removed the `Pickable` component, since until now, it serves no purpose. It may be reinstantiated.

I've also done some refactoring in the `assets.rs` module, consistently using the convenience loading API that was used only a few times.

This PR addresses the first entry in #83.


64kramsystem:
@zicklag Do you know why the AssetLoader is not invoked at all in this branch ([here](https://github.com/64kramsystem/punchy-dev/blob/spawn_items/src/assets.rs#L270))?

I've put everything in place, but assets under /items/ are ignored. I've tried to figure out what triggers a load, but looking at fighters for example, doesn't seems to lead to any explicit loading code (so I take there's an implicit load somewhere).

zicklag:
You need to load them as a part of the level load with another block like this one:

https://github.com/64kramsystem/punchy-dev/blob/ae35d77424e3885e8ddd045f7691335224738ecf/src/assets.rs#L169-L177

64kramsystem:
> You need to load them as a part of the level load with another block like this one:
> 
> https://github.com/64kramsystem/punchy-dev/blob/ae35d77424e3885e8ddd045f7691335224738ecf/src/assets.rs#L169-L177

Thanks! I'm actually already doing that [here](https://github.com/64kramsystem/punchy-dev/blob/spawn_items/src/main.rs#L363).

[Here](https://github.com/64kramsystem/punchy-dev/blob/spawn_items/src/main.rs#L414) I try to load the `ItemMeta`, but nothing is found.

I wonder if I'm using the wrong handle or something :thinking:

64kramsystem:
Ah! I need to do all the manual load :flushed: I'll try this, thanks :smile:

zicklag:
> Thanks! I'm actually already doing that [here](https://github.com/64kramsystem/punchy-dev/blob/spawn_items/src/main.rs#L363).

No, you have to do it in the Level _asset loader_. The level asset loader is responsible for telling the asset server that it depends on the item assets, and for setting the item handles to their actual handles, otherwise they will default to null handles that won't spawn anything.

64kramsystem:
@erlend-sh CC

zicklag (approved):
Looks great! 

I like your attention to code style and quality and improvement. :+1:

PR #123, opened 2022-07-23
Weapons refactoring

64kramsystem:
(name of the branch is misleading; this is because I started refactoring while I was writing the item picking code...)

This refactoring is a preliminary step to implementing items picking.

## Main refactorings

The first has been to better name the weapon-related concepts:

- the player's weapon attacks were named "attack" and "throw", which made them confusing; now, the first is named "shoot"
- the Item component has been renamed to Weapon ("Item" will describe objects that can be picked)

The second has been to make bundles explicit, by creating ThrownWeaponBundle and ShotWeaponBundle. While this may seem optional, it gives a clear and immediate idea of the base components of a bundle.

With refactoring 2, it came out the fact that the weapon attacks code was almost-duplicate, and that the shoot attack was not using Item (now Weapon). So the third refactoring has been to merge the related code.

## Other refactorings

- Removed the two (event-based) item throwing systems into one; event-based was ok, but the shooting system was not event-based, so it was best to make them consistent
- Moved some other systems into the attack plugin (module), which now better encapsulates the attack-related logic

## Considerations

Since the program can only grow in complexity, it gets progressively more important to keep a consistent structure.

For example, it took me a long while (and hair :exploding_head:) to realize that there were two separate weapon collision systems, because the naming of `item_attacks_enemy_collision` hinted that it was the only one.

Two helpful strategies for the future are:

1. move the remaining code out of main; with clear modules, it's clearer to find and understand conceptually related code, and correctly add new ones
2. explicitly defining bundles probably helps with understanding their structure, and operate on them (e.g. comparing them)


zicklag:
Awesome! I'll be a bit busy today and tomorrow, so I'll probably get to reviewing this and your other PRs on Monday and there's no problem merging before I get to it if @odecay checks it out.

64kramsystem:
Rebased.

odecay (approved):
Looks good, gonna get this in as it is touching code I'm working on as well and I want to rebase.

PR #126, opened 2022-07-24
Add items picking logic

### LINKED ISSUE #83: Bottle throwable item

64kramsystem:
Add items picking logic:

- items picked are associated to the player
- bottles are thrown/shot only after being grabbed
- after throwin/shooting a bottle, the item is not carried anymore

A minor but subtle bug is also fixes see 7e22c45eb9b33f0b08c447e4c81acc4f7031fed0.

Also note the side effect described in d7cdbbf5b30cd5ab276a02d7877830117dcd57d3.

Closes #83.


64kramsystem:
Rebased.

src/item.rs
```diff
@@ -1,33 +1,108 @@
...
 
+#[derive(Component)]
+pub struct Item;
+
+#[derive(Component)]
+pub struct CarriedBy(pub Entity);
+
+/// Represents an item, that is either on the map (waiting to be picked up), or carried.
+/// If an item is on the map, it has a TransformBundle; if it's carried, it instead has a
+/// CarriedBy.
```
odecay, line 27: Im wondering if instead of ridding the item of its `TransformBundle` we might consider pushing it as a child of the Fighter who picks it up?

odecay (approved):

src/item.rs
64kramsystem: Make sense!

odecay: It might be awkward at first before we figure out how to integrate it with animations/what offset/rotation and whatnot it should have, but for now we could always toggle the visibility of the item off when its picked up as well.

zicklag (approved):
I'm lacking bandwidth to do a total review right now, but looks fine after a quick sanity check and it seemed to work after testing it out real quick.

src/item.rs
64kramsystem: There you go! Done. It actually made the code a little bit simpler :smile:

64kramsystem: Ok. There have been several conflicts, that I had to fix. I've verified the rebase, hopefully, there have been no subtle breakages :sweat_smile:.

src/item.rs
```diff
@@ -1,33 +1,102 @@
...
+            // If several items are at pick distance, an arbitrary one is picked.
+            for (item_id, item_transform) in items_query.iter() {
+                if !picked_item_ids.contains(&item_id) {
+                    let player_item_distance = player_transform
+                        .translation
+                        .truncate()
+                        .distance(item_transform.translation.truncate());
+
+                    if player_item_distance <= PICK_ITEM_RADIUS {
+                        commands.entity(item_id).remove_bundle::<TransformBundle>();
```
odecay, line 73: I meant just to keep the transform bundle either way but I guess this is fine as well, we can always change it later if we need to use the transform for animations in the future.

64kramsystem: ok. if we put it in the hads or something, then it will certainly go back :smile: it is unnecessary now, since if we put it in the HUD (there is a tracking issue), the HUD uses different semnantics/types anyway.

64kramsystem:
Ok. Ready for review and hopefully merge. Implemented the change requested in one comment, and resolved the other comment without change.

odecay (approved):
:+1:

odecay:
bors merge

Issue #99, opened 2022-07-16
Solve commands flushing (1-frame-bug related) issues


odecay:
EDIT: #203 was made to solve (more or less completely) this problem. Currently, the bug mentioned below is still present (still requires a workaround).

Currently, we have commands flushing issues (1-frame-lag related bugs), and likely race conditions.

Structuring commands flushes is something that IMO needs to be tackled ASAP, because the more time passes, the more (and more subtle) bugs will arise.

Currently the game has [one recognized bug](https://github.com/fishfight/punchy/blob/72c83a68b1e931f470b38c3513d6eeef80b86757/src/attack.rs#L363), which requires a workaround. I suspect that there are also race conditions.

Note that this is orthogonal to state design.

64kramsystem:
I totally agree with this proposal. :joy:

zicklag:
Seems reasonable to me.

64kramsystem:
@zicklag @odecay Phew! :sweat_smile: let me know what you think.

odecay:
My initial impressions are of this are very positive. I feel like what you have done in sorting through all the systems already to list them like this was a big task, and the way you have ordered them makes perfect sense to me.

I will think about this some more and see what other feedback I can suggest but I feel this is a good way to proceed.

64kramsystem:
> In regards to events, my understanding of what you are saying is that you find them helpful but there will be less need to rely on them in this architecture.

Yes, exactly. Event messaging, in this context, can serve two purposes:

1. to establish a temporal association - a reader implicitly acts _after_ a writer, and/or to establish a 1:1 association between systems (when there is only 1 publisher and 1 subscriber)
2. to model a single publisher/multiple subscribers model

Point 2 is what event messaging is exactly for, so they're an exact fit :smile: 

Regarding Point 1, with proper stages in place, if one wants to model temporal association, they won't need events (or labels) anymore, so they can model via components if they wish. Using events for 1:1 associations will be therefore a more cosmetic choice, which is of course a valid choice.  
We may decide to establish a guideline, so that we have a consistent usage; a decision to model 1:1 associations via event messaging (and later change an association to be components-based if it becomes 1:many) is perfectly valid.

zicklag:
I haven't done a thorough review, but I think I'll leave that mostly to you guys so I keep my focus on the non-gameplay stuff I've been focusing on. :) And good job on the breakdown!

[...]

Issue #54, opened 2022-07-06
Define attack startup frames


odecay:
Attacks can have a few frames of windup before spawning an active hitbox #53 . This allows for players and enemies to be hit out of an attacks windup animation.

Issue #53, opened 2022-07-06
Attack Hitbox update


odecay:
Collision should be changed so that attacks have hitboxes and fighters have hurtboxes. This will allow us to decide what to do when attacks collide with one another.

![ErNNP](https://user-images.githubusercontent.com/1379590/177467750-1a92b848-db51-4ff9-9cfd-c7aa8602fc8f.jpg)
Area in red is the active hitbox of the attack,  Area in green is the hurtbox of the player during the attack. If the attacks hitbox connects with a players hurtbox the enemy will take damage, if the attack overlaps with an enemies active hitbox, attacks could clash.

PR #134, opened 2022-07-25
Attack hitbox and startup frame refactor

odecay:
Closes both
#53 #54 

Collisions have been refactored to take place in one system only, the filtering we need is already done via `BodyLayers`.
There are now two main types of attacks.

**1.  Melee attacks:** these are spawned as a child of the fighter who starts them, with an `AttackFrames` component which consists of the startup active and recovery frames of the attack. These correspond to the frame window in which a hitbox is spawned, which will collide with other fighters as defined by its collision layers.  Melee attacks are not despawned when they collide with something, instead they stay active for the duration of their Active Frames. There are systems which activate the hitbox at the start of the Active Frames and despawn the attack at the end of the Active Frames/Start of the Recovery Frames. These attacks also lock the fighter into the `Attacking` state for the duration of their attack animation.

**2. Projectile attacks:** these are spawned un-parented, and with a `ProjectileLifetime` component. They are despawned on collision, or upon reaching the end of their lifetime. Currently only the player has these, and they do not lock the player into an attack animation, although I think ultimately both of these things will change, players and enemies will both want projectile attacks and they will be tied to either character specific special attacks or Ranged Weapon animations.

### Things to consider for this PR:

- Currently there may be issues with the exactness of frame timing due to commands being used to activate hitboxes and the need for commands to be separated into stages to apply. This should probably be addressed after #99, but I am confident that these frame based definitions for attacks are the correct design. I tried the approach of changing adding the hitbox component at attack spawn time with a 0 sized shape, and mutating to the correct shape on Active frame start to avoid command application, but this did not behave as expected.
- I may think further about the startup, active, and recovery attributes of the `AttackFrames` component. I want to stick with these names but I see the way they are referenced currently could be confusing, currently they are the index relative to the animation on which the startup, active, and recovery portions of the attack **_end_**. I also considered representing them as ranges, but it seemed like an index was simpler.
- I have refactored some of the work done in #123 and that lead to some renames from what was added by @64kramsystem. I think I will take another pass tomorrow to check but I am quite confident that the conceptual model of Melee and Projectile attacks is cleaner for our purposes here. (Some of what that rework took into account was due to the way flop worked before but it is now unified with other Melee attacks)
- I have left some unnecessary comments which I will come back and clean up before merge.

### Further improvements

- [ ] Define attack startup frames, hitbox size and placement via transform relative to parent in fighter yaml files.
- [ ] Write up further design goals and spec for attacks.
(I intend to do both of these)


zicklag (approved):
I did a quick, light review and it looks good to me.

I like that we've moving things into a more generic Projectile/Melee attack systems with less item specific systems. :+1:

src/attack.rs
```diff
@@ -343,15 +373,53 @@ fn enemy_attack(
...
+}
+
+fn deactivate_hitbox(
+    query: Query<(Entity, &AttackFrames, &Parent), (With<Attack>, With<Collider>)>,
+    fighter_query: Query<&Animation, With<State>>,
+    mut commands: Commands,
+) {
+    for (entity, attack_frames, parent) in query.iter() {
+        if let Ok(animation) = fighter_query.get(parent.0) {
+            if animation.current_frame == attack_frames.recovery {
```
64kramsystem, line 402: In general design terms, should we account for lost frames (as a consequence, avoid exact matches where possible)? They're edge cases, so I don't know if it's worth or not.

src/attack.rs
```diff
@@ -256,16 +265,37 @@ fn player_flop(
...
+                    .insert(AttackFrames {
+                        startup: 0,
+                        active: 3,
+                        recovery: 4,
+                    })
+                    .id();
+                commands.entity(entity).push_children(&[attack_entity]);
+                //TODO: define hitbox size and placement through resources
+
+                //maybe move audio effects?
```
64kramsystem, line 287: Possibly! It depends on what you have in mind :smile: 

Maybe adding a system that matches the current on-screen animations? I haven't checked the system, in this case, they'd need to be connected to a state.

64kramsystem (approved):
Looks good to me :smile:

src/attack.rs
odecay: I agree that makes sense but also I would like to figure out a way to not have these sort of lost frames at all. In this case ill just replace this with `>=`.

src/attack.rs
odecay: Not sure I have anything in mind yet, just noting that it feels a bit awkward that it lives here, I think this might make sense to either trigger based off attack actionlike state or when we split out the movement intent from attacks also use a similar strategy to split out the sound triggering intent of attacks.

64kramsystem: Difficult to say :thinking: The reason is that this is already a decoupled logic - `FighterStateEffectsPlayback` does not play any effect itself, but it adds a component, in turn this component is picked up by a separate system (`fighter_sound_effect`).

Having said that, I definitely find ugly the surrounding conditionals. They can be offloaded to the `fighter_sound_effect` system, by passing the `Handle<FighterMeta>`, so that the audio system can find out the asset by itself. On the other hand, this design leads to one system is saying to another "here's the metadata; maybe you'll play a sound, maybe you won't". This can be fine, it's a matter of taste.

Let me know if you have any preference, since I think all the approaches are equally valid.

odecay:
bors +r

odecay:
> Did you mean "r+"?

Did I ? :laughing:

odecay:
bors merge

zicklag:
Haha, check it out, bors actually merged both of our pull requests into a single merge commit because we tried to merge it at the same time. That's cool. :)

PR #142, opened 2022-07-25
Updates to Github workflows, primarly remove cache building and leave build to release

### LINKED ISSUE #138: Slim down project build in Github workflows
### Discussed in https://github.com/fishfight/punchy/discussions/136

<div type='discussions-op-text'>

<sup>Originally posted by **zicklag** July 25, 2022</sup>
For some reason, maybe just because we have a lot of dependencies, our builds are taking a pretty long time. 13 minutes for the Linux build, thought we somehow got 13 seconds with the Mac build ).

It's probably because we're running out of cache space and it just starts deleting some of our caches resulting in the cached builds running really fast and the un-cached builds taking the hit.

Anyway, I just realized, though, why do we actually need to run a full build for CI? Can't we just `cargo check` ( or better yet `cargo clippy` ) on all of the target backends instead? We're not going to use the built artifacts for anything from CI. I can't think of any real code problems that that wouldn't detect other than linker errors.

There are some crates/tools that cause linker errors to prevent you from doing certain things, but those are usually hacky workarounds for something like a way to prevent you from inserting unsafe code into a function. If we're not using anything like that, I think a cargo check in CI should be just about as good as a cargo build. Then we only need to do actual cargo builds on releases.

There might be a scenario I'm not thinking of. What do you think? @odecay @64kramsystem @orhun</div>

64kramsystem response:

> I'm not 100% sure why builds are so slow, as 10 GB should be enough, however, given that there are no test suites in the project, bulding only for tags makes sense.
>
> Also, now that the CI workflow builds also on master, we don't need the Build caches workflow anymore, so it's additional resources freed.

64kramsystem:
- Remove build caches workflow
  - Minor updates to workflows, mostly cosmetic
  - Update checkout action version
  - Remove apt-get --allow-unauthenticated option, which is not relevant in this context
  - Change apt-get mode from very quiet to quiet, since some messages may be useful (e.g. package already installed)
  - Made the two dependency installation jobs the same
- Make tags pattern consistent across cd and web-dem-
- Remove project building from CI (left in CD)

Closes #138.


odecay:
I don't have any meaningful review on this but I want to get it in asap :+1:

.github/workflows/ci.yml
```diff
@@ -71,14 +39,7 @@ jobs:
...
-          toolchain: stable
-          override: true
-          profile: minimal
-          target: ${{ matrix.config.target }}
       - uses: actions-rs/cargo@v1
         with:
-          command: build
-          args: --locked --target ${{ matrix.config.target }}
+          command: clippy
+          args: -- -W clippy::correctness -D warnings
```
zicklag, line 45: I think we still want to run the clippy check as a matrix on different platforms to capture differences in the build path on different platforms.

It may not be necessary for different desktop platforms, though. Maybe just web and native. It depends on whether or not we end up adding any OS specific doe into Punchy itself. I think since Bevy handles the abstraction over desktop platforms, we'd be good just checking Linux and web.

zicklag (changes requested):

.github/workflows/ci.yml
zicklag: Oh, and maybe we should add with and without the `--features debug` flag to the matrix, too. @odecay had mentioned the debug build kept failing because we don't tend to test it out as much.

64kramsystem: I think we could make those changes in a separate PR? So that we have more time to think them through.

zicklag: Fair enough.

zicklag (approved):

zicklag:
bors merge

zicklag:
bors merge

PR #143, opened 2022-07-25
Fix World Cleanup and Refactor as System Param

### LINKED ISSUE #131: Exiting Pause menu behaves unexpectedly
### Description

Exiting pause menu returns the player to the main menu, abandoning in progress gameplay.
This is not the desired behavior either for gameplay or debugging purposes. Ideally for gameplay exiting pause state would return the players to in progress session.
For debugging it is useful to not have the current world despawned while the pause is taking place, so that entities and components can be inspected using bevy-inspector-egui.

### To Reproduce

Pause the game, exit pause menu by pressing ESC key or Continue menu button.

zicklag:
Fixes #131

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -46,6 +46,7 @@ mod platform;
 mod player;
 mod state;
 mod ui;
+mod utils;
 mod y_sort;
 
 use animation::*;
@@ -58,6 +59,7 @@ use movement::*;
 use serde::Deserialize;
 use state::{State, StatePlugin};
 use ui::UIPlugin;
+use utils::ResetController;
 use y_sort::*;
 
 use crate::{config::EngineConfig, input::PlayerAction, item::ItemSpawnBundle, metadata::Settings};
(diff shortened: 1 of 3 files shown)


64kramsystem (approved):
:+1:

src/main.rs
```diff
@@ -211,7 +213,6 @@ fn main() {
         .add_startup_system(set_audio_channels_volume)
         .add_enter_system(GameState::InGame, play_level_music)
         .add_exit_system(GameState::InGame, stop_level_music)
-        .add_exit_system(GameState::InGame, clean_in_game_data)
```
odecay, line 214: Is this the change which basically prevents cleanup from happening everytime `GameState::InGame` exits? This is the fix and the rest is mostly just cleanup correct?

odecay (approved):

src/main.rs
zicklag: Yep, this is the fix.

The rest of the code just makes the logic re-usable.

odecay: Cool let's get it in

zicklag:
bors merge

Issue #140, opened 2022-07-25
MVP Scripting 


zicklag:
### Description

We want to get an initial pass working on a scripting API. The goal here isn't to produce any final quality solution but to get the ball rolling and start finding out what the challenges we are going to run into are going to be.

I think a good goal is to allow you to do something useful in the game with a script. This could be anything, but I'm leaning towards an item implementation. For instance, maybe we try to migrate the implementation of the bottle item to a script.

Let's try to allow you to control things such as the bottle's flight pattern and damage.

---

As we implement, I think we want to try to keep the following ( future? ) goals in mind:

- It'd be good for scripts to be usable as kind of a dynamic alternative to the [Fish Fight Jumpy Item JSONs](https://github.com/fishfight/FishFight/tree/main/assets/items). Instead of having static JSON, lets try to make scripts as flexible as reasonable so you are less dependent on the item options that you have, for example, in Fish Fight.
- We don't have to let you do _everything_ in scripts. Doing a total and absolute binding to everything in Bevy isn't really necessary for this game. We want to remove as many barriers in scripting as possible, but focus on making it possible to modify _content_ with scripts, not game architecture.
- We want to give item implementations as much read access to relevant info in the game world as we can. For instance, an item should be able to use info such as where the enemies are, where the players are, what items and obstacles are where on the map, how much life fighters have, etc. Scripts can use this info for all kinds of interesting things, such as a homing weapon that targets the fighter with the most life or similar effects.

### Alternatives & Prior Art

- https://github.com/fishfight/punchy/issues/107

odecay:
I'm not familiar with the Jumpy item JSONs but I was thinking as we get closer to having different types of weapons and attacks, it may be nice to script behaviors for them in their resource files.

zicklag:
That's what I was thinking.

In Jumpy they have a rather clever setup of effect spawners, projectiles, and collision boxes that can all be defined in JSON and create a rather flexible arrangement of weapon behaviors, but it's still all static in JSON and restricted to the options exposed by the game JSON metadata. It's similar to how all of our stuff is defined in YAML today.

We'd want something similar, but using scripts instead for applicable assets such as items, so that you aren't as restricted in what kind of item behaviors you can add.

If possible, I want to try to push as much of the content related behaviors as makes sense into scripts. So items and fighter attacks ( and possibly movement ) could be totally customized with mods or in other games on the "Punchy engine".

zicklag:
[...]

PR #147, opened 2022-07-25
Implement First Pass Script Execution

zicklag:
- [x] Added .editorconfig file to configure different tab sizes for JavaScript and Rust. This helps for editors with editorconfig enabled to give a consistent style. Without it things can get weird because of editor defaults for different languages not complying with installed formatters.
- [x] Added TypeScript and JavaScript asset loaders
- [x] Added script loading and hot reloading
- [x] Added script log API
- [x] Rework script loader to use modules, i.e. `export` and `import`.
- [x] Rework to not use modules again.
- [x] Implement scripting engine for WASM backend
- [ ] Implement component modification by scripts

Works towards #140.

Not done yet, but in a rather stable state that is very indicative of the direction I'm going with it.


zicklag:
Closing this now that my changes have been merged into `bevy_mod_js_scripting`.

Issue #121, opened 2022-07-23
(Main menu) Music is not stopped, when the start/stop cycle is very short (autostart flag)


64kramsystem:
### Description

When the autostart flag is set, the main menu music playback is not stopped, resulting in both the main menu and the level tracks being played.

I've tried the patch below, which gives a few frames of delay before the stop invocation, but the problem persists. 

My educated guess is that the audio plugin has a small delay before it starts playing a given music/sound, and if a stop command is issued in the meanwhile, it's ignored.

EDIT: It seems that this is a [known Bevy Kira audio issue](https://github.com/NiklasEi/bevy_kira_audio/issues/62).

Patch (against e183020e27):

```diff
diff --git i/src/audio.rs w/src/audio.rs
index 5918fea..634213f 100644
--- i/src/audio.rs
+++ w/src/audio.rs
@@ -76,11 +76,8 @@ pub fn fighter_sound_effect(
 }
 
 pub fn play_menu_music(game_meta: Res<GameMeta>, music_channel: Res<AudioChannel<MusicChannel>>) {
-    music_channel.play(game_meta.main_menu.music_handle.clone());
-}
-
-pub fn stop_menu_music(music_channel: Res<AudioChannel<MusicChannel>>) {
     music_channel.stop();
+    music_channel.play(game_meta.main_menu.music_handle.clone());
 }
 
 pub fn play_level_music(
@@ -88,11 +85,8 @@ pub fn play_level_music(
     assets: Res<Assets<LevelMeta>>,
     music_channel: Res<AudioChannel<MusicChannel>>,
 ) {
+    music_channel.stop();
     if let Some(level) = assets.get(level_handle.clone_weak()) {
         music_channel.play(level.music_handle.clone());
     }
 }
-
-pub fn stop_level_music(music_channel: Res<AudioChannel<MusicChannel>>) {
-    music_channel.stop();
-}
diff --git i/src/main.rs w/src/main.rs
index 9a81a72..9f73773 100644
--- i/src/main.rs
+++ w/src/main.rs
@@ -217,7 +217,6 @@ fn main() {
         .add_system(platform::load_storage.run_in_state(GameState::LoadingStorage))
         .add_startup_system(set_audio_channels_volume)
         .add_enter_system(GameState::InGame, play_level_music)
-        .add_exit_system(GameState::InGame, stop_level_music)
         .add_exit_system(GameState::InGame, clean_in_game_data)
[...]

PR #151, opened 2022-07-26
issue 121 is already closed

DRuppFv:
Comment is no longer needed. ~ and issue 121 is already closed

### DIFF
--- a/src/audio.rs
+++ b/src/audio.rs
@@ -81,12 +81,6 @@ pub fn play_menu_music(
     music_channel: Res<AudioChannel<MusicChannel>>,
     engine_config: Res<EngineConfig>,
 ) {
-    // When the autostart flag is set, the main menu music playback is not stopped, resulting in both the main menu and the level tracks being played.
-    // I've to introduce a few frames of delay before the stop invocation, but the problem persists.
-    // My educated guess is that the audio plugin has a small delay before it starts playing a given music/sound, and if a stop command is issued in the meanwhile, it's ignored.
-    //
-    // See issue #121.
-    //
     if !engine_config.auto_start {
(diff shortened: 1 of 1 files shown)


src/audio.rs
```diff
@@ -81,12 +81,6 @@ pub fn play_menu_music(
     music_channel: Res<AudioChannel<MusicChannel>>,
     engine_config: Res<EngineConfig>,
 ) {
-    // When the autostart flag is set, the main menu music playback is not stopped, resulting in both the main menu and the level tracks being played.
-    // I've to introduce a few frames of delay before the stop invocation, but the problem persists.
-    // My educated guess is that the audio plugin has a small delay before it starts playing a given music/sound, and if a stop command is issued in the meanwhile, it's ignored.
-    //
-    // See issue #121.
-    //
     if !engine_config.auto_start {
```
zicklag, line 84: While #121 is already closed, the comment was meant to justify the need for checking the autoplay flag right there, because casual reasoning could lead you to believe that the check was unnecessary.

Maybe we just make that more clear here.

```suggestion
    // This is a workaround for a Bevy Kira bug where stopping a sound immediately after
    // playing it doesn't work. We run into this issue when the menu starts and immediately
    // stops because the auto-start flag skips the menu. See issue #121 for context.
    if !engine_config.auto_start {
```

zicklag (changes requested):

DRuppFv:
@zicklag You requested changed but the changes were already committed by you. What should i do then? Just wait for the merge?

DRuppFv:
don't mind me closing by accident :rofl:

zicklag:
Yeah, I'll just merge it now. Thanks for opening. :) 

bors merge

PR #154, opened 2022-07-26
Implement player movements clamping (currently, via helper function)

### LINKED ISSUE #63: Unify (if possible) the player translation (and limits)

### LINKED ISSUE #133: Possible to flop past stage boundary
### Description

Flopping at the X axis stage boundary allows the player to move past it, subsequently becoming unable to move other than continuing to flop.

Fix this during #63.

64kramsystem:
Clamp the player movements/attacks, currently via function. Once the multistage design is implemented, this can be easily split (which will actually simplify the code).

I'm marking this as closes #63 and closes #133, as this effectively fixes the problem. Once the multistage is in place, intention to move is essentially a refactoring.

### DIFF
--- a/src/attack.rs
+++ b/src/attack.rs
@@ -25,8 +25,10 @@ use crate::{
     },
     input::PlayerAction,
     item::item_carried_by_player,
-    metadata::{FighterMeta, ItemMeta},
-    movement::{MoveInArc, MoveInDirection, Rotate, Target},
+    metadata::{FighterMeta, GameMeta, ItemMeta},
+    movement::{
+        clamp_player_movements, LeftMovementBoundary, MoveInArc, MoveInDirection, Rotate, Target,
+    },
     state::State,
     ArrivedEvent, Enemy, GameState, Player,
 };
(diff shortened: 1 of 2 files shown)


64kramsystem:
Curses! I need to solve the conflicts :exploding_head:

odecay (commented):
Code looks good to me, going to test it out quickly.

Edit: Works great too.

odecay:
bors merge

64kramsystem:
Yay, we have clamping :star_struck:

PR #159, opened 2022-07-27
Add more enemies

edgarssilva:
The image property, of the fighter yaml files, now receives an array of image paths that will be loaded and stored.
When adding the correct components for the enemy a random texture atlas is picked.


edgarssilva:
Outdated branch

PR #166, opened 2022-07-27
temp fix for attack frames

odecay:
Spritesheet changing animation frames caused breakage with hitbox despawning on attacks fixed here.


odecay:
bors merge

PR #168, opened 2022-07-28
Remove Strip Flag in Release Profile

### LINKED ISSUE #167: WASM Release Build Does Nothing
### Description

Making a release build for WASM will render the loading screen endlessly.

### To Reproduce

1. `just build-release-web`
2. `cd target/wasm-dist`
3. `basic-http-server -x .`
4. Open `http://localhost:4000` in your browser

### Expected Behavior

The web page should load

### Additional Context

The issue appears to be related to compiler optimizations. Some observations so far:

- When I build with default release opt-level and debug_assertions = true it works
- When I build a release with default options, it doesn't work
- When it doesn't work, it doesn't do anything at all in Rust land it seems. Even putting a panic at the top of `main()` doesn't get a response.
- One time, I can't remember how exactly, I was able to get it to throw an error that it was reaching an `unreachable!()` somewhere.

### Log Messages

_No response_

zicklag:
This was breaking WASM release builds. Also I think we may want to leave
it out anyway, because stripping removes panic info that we probably
want, at least this early in development.

Fixes #167

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,7 +56,6 @@ debug = false
 # slower than opt-level=0.
 
 [profile.release]
-strip = true
 lto = true
 codegen-units = 1 # Improves physics performance for release builds
 


zicklag:
bors merge

PR #170, opened 2022-07-28
Web Build & CI Improvements

zicklag:
- Make web build use it's own target folder to streamline local
development of both native and web builds.
- Make web build URL path agnostic so it can be hosted under any path.
- Make web player build in CI use different paths for each release along
with a latest path that symlinks to the latest release.

### DIFF
--- a/.github/workflows/web-demo.yml
+++ b/.github/workflows/web-demo.yml
@@ -3,7 +3,6 @@ name: Publish Web Demo
 on:
   push:
     tags:
-      - web-demo
       - 'v*'
 jobs:
   publish-web-demo:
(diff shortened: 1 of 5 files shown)


odecay (approved):

zicklag:
As soon as this build finishes we'll find out if this works.

https://github.com/zicklag/punchy/runs/7564619642?check_suite_focus=true

zicklag:
Yay it worked!

That gave us a tag specific build: https://zicklag.github.io/punchy/player/v0.0.3-rc2/

And automatically updated the latest symlink to point to the latest build: https://zicklag.github.io/punchy/player/latest/

This should be good to merge then.

---

As a note, after we make the release and tag, we will want to manually push a commit to the `gh-pages` branch to move the current `/demo` folder to `/player/v0.0.2`.

zicklag:
bors merge

PR #172, opened 2022-07-29
Read Player Attack Damage From Stats

zicklag:
Player attack damage was hard-coded to 10.


64kramsystem (approved):

64kramsystem:
uh, looks like cargo fmt is complaining :flushed:

PR #173, opened 2022-07-29
Fix Bug With Hot Reload and New Loading Progress

zicklag:
PR #164 broke hot reloading by accidentally triggering asset update
events every time it would check the game load progress.

This moves game load progress checking to a system condition instead
with a read-only borrow to the game asset to prevent the issue.


zicklag:
Ah, for some reason the condition system is running every frame no matter what and logging loading progress. I'll fix that.

zicklag:
OK, that should fix it.

64kramsystem (approved):

PR #174, opened 2022-07-29
Implement Enemy FOV

64kramsystem:
Very simple, based on a fixed X distance.

### DIFF
--- a/src/consts.rs
+++ b/src/consts.rs
@@ -7,6 +7,9 @@ pub const PLAYER_HEIGHT: f32 = PLAYER_SPRITE_HEIGHT - 50.;
 //
 pub const LEFT_BOUNDARY_MAX_DISTANCE: f32 = 380.;
 
+// In this style of game, considering Y is not significant.
+pub const ENEMY_FOV_X: f32 = 800.;
+
 pub const GROUND_Y: f32 = -120.;
 pub const GROUND_HEIGHT: f32 = 150.;
 pub const GROUND_OFFSET: f32 = 0.;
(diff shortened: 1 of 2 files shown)


src/consts.rs
```diff
@@ -7,6 +7,9 @@ pub const PLAYER_HEIGHT: f32 = PLAYER_SPRITE_HEIGHT - 50.;
 //
 pub const LEFT_BOUNDARY_MAX_DISTANCE: f32 = 380.;
 
+// In this style of game, considering Y is not significant.
+pub const ENEMY_FOV_X: f32 = 800.;
```
zicklag, line 11: Just a thought, should this be an enemy stat, or not really?

Anyway, not something that needs to be done for this PR.

zicklag (approved):
:+1:

src/consts.rs
64kramsystem: I think the exact solution is tripping points (so, level metadata).

As far as I remember from Final Fight and similar, as long as the player was inside the screen, they were always chased by on-screen enemies, so the FOV was effectively one screen. However, I think beat'em ups don't work with FOV actually, rather, with tripping points. FOV doesn't work well though, because if there is a camera stop (wave to beat), the next wave of enemies should be placed at least ENEMY_FOV_X (which is currently one screen), which may be too far.

So this is essentially a stopgap solution to avoid enemies running at the player from the whole level :smile:.

The next step, depending on priorities, will be to add tripping points to the level meta.

zicklag: That makes sense. I'll call this resolved, then and go ahead and merge it.

64kramsystem:
I definitely need to play Final Fight again. I never though I'd end up working on a beat 'em up :joy:

PR #178, opened 2022-07-30
Move EngineConfig to a Static Global

zicklag:
EngineConfig is read-only and logically makes sense as a static as it
last the life of the program, so this moves it to a global static that
can be accessed from anywhere in the program instead of just in the Bevy
world.

This is also going to be required for scripting support so that the
asset directory will be readable from asset loaders without borrowing the
Bevy world.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3170,6 +3170,7 @@ dependencies = [
  "getrandom",
  "iyes_loopless",
  "leafwing_input_manager",
+ "once_cell",
  "punchy_macros",
  "rand",
  "serde",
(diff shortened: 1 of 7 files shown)


64kramsystem (approved):
LGTM :smile: (there's a typo (`EGNINE`) :smile:)

Once Rust 1.64 is released, I think that we can get rid of `once_cell`, since Mutex will be const :heart_eyes:

zicklag:
Wow. :facepalm: Thanks for catching the typo. :)

> Once Rust 1.64 is released, I think that we can get rid of once_cell, since Mutex will be const heart_eyes

Actually we don't need mutex, though, we just need the Lazy functionality to initialize the value on first access. But eventually when [`Lazy`](https://doc.rust-lang.org/stable/std/lazy/struct.Lazy.html) is stabilized, then we'd be good. :)

PR #181, opened 2022-07-30
Convert (InGame) design to full multi-stage

### LINKED ISSUE #99: Solve commands flushing (1-frame-bug related) issues

64kramsystem:
Create a (InGame) design fully based on Bevy stages, for the reasons explained in #99.

This fixed at least on bug! See 7e9e854815.

Note that there is an existing issue, which is unrelated to this PR - if the flop is executed at the top location, the player will only go down.

I've tested this change for minutes 10, maybe more, with different cycles, and didn't find any issue, but unfortunately, this type of change may introduce subtle bugs.

I think that the stages introduced are consistent; the "Processing" name is something I find a bit too generic (but I couldn't find any better). Of course the PR is open to changes 😄.

Closes #99.

### DIFF
--- a/src/attack.rs
+++ b/src/attack.rs
@@ -47,7 +47,8 @@ impl Plugin for AttackPlugin {
                 .with_system(enemy_attack)
                 .into(),
         )
-        .add_system_set(
+        .add_system_set_to_stage(
+            GameStage::Processing,
             ConditionSet::new()
                 .run_in_state(GameState::InGame)
                 .with_system(activate_hitbox)
--- a/src/attack.rs
+++ b/src/attack.rs
@@ -6,7 +6,7 @@ use bevy::{
     math::{Vec2, Vec3},
     prelude::{
         default, App, AssetServer, Assets, Bundle, Commands, Component, Entity, EventReader,
-        Handle, Local, Parent, Plugin, Query, Res, Transform, With, Without,
+        Handle, Parent, Plugin, Query, Res, Transform, With, Without,
     },
     sprite::SpriteBundle,
     transform::TransformBundle,
@@ -303,7 +303,6 @@ fn player_flop(
     time: Res<Time>,
     left_movement_boundary: Res<LeftMovementBoundary>,
     game_meta: Res<GameMeta>,
-    mut start_y: Local<Option<f32>>,
 ) {
     let players_movement = query
         .iter_mut()
@@ -365,22 +364,10 @@ fn player_flop(
                         }
                     }
 
-                    // For currently unclear reasons, the first Animation frame may run for less Bevy frames
-                    // than expected. When this is the case, the player jumps less then it should, netting,
-                    // at the end of the animation, a slightly negative Y than the beginning, which causes
-                    // problems. This is a workaround.
-                    //
-                    if start_y.is_none() {
-                        *start_y = Some(transform.translation.y);
-                    }
-
                     if animation.current_frame < 1 {
                         movement.y += 180. * time.delta_seconds();
                     } else if animation.current_frame < 3 {
                         movement.y -= 90. * time.delta_seconds();
-                    } else if animation.is_finished() {
-                        movement.y = start_y.unwrap();
-                        *start_y = None;
                     }
 
                     (transform.translation, Some(movement))
--- a/src/attack.rs
+++ b/src/attack.rs
@@ -30,29 +30,31 @@ use crate::{
         clamp_player_movements, LeftMovementBoundary, MoveInArc, MoveInDirection, Rotate, Target,
     },
     state::State,
-    ArrivedEvent, Enemy, GameState, Player, Stats,
+    ArrivedEvent, Enemy, GameStage, GameState, Player, Stats,
 };
 
 pub struct AttackPlugin;
 
 impl Plugin for AttackPlugin {
     fn build(&self, app: &mut App) {
-        app.add_system_set(
+        app.add_system_set_to_stage(
+            GameStage::Actions,
             ConditionSet::new()
                 .run_in_state(GameState::InGame)
                 .with_system(player_projectile_attack)
                 .with_system(player_throw)
                 .with_system(player_flop)
+                .with_system(enemy_attack)
+                .into(),
+        )
+        .add_system_set(
+            ConditionSet::new()
+                .run_in_state(GameState::InGame)
                 .with_system(activate_hitbox)
                 .with_system(deactivate_hitbox)
                 .with_system(projectile_cleanup)
                 .with_system(projectile_tick)
                 .into(),
-        )
-        .add_system(
-            enemy_attack
-                .run_in_state(GameState::InGame)
-                .after("move_to_target"),
         );
     }
 }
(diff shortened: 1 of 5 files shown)


src/attack.rs
```diff
@@ -30,29 +30,32 @@ use crate::{
...
                 .with_system(activate_hitbox)
                 .with_system(deactivate_hitbox)
                 .with_system(projectile_cleanup)
                 .with_system(projectile_tick)
                 .into(),
-        )
-        .add_system(
-            enemy_attack
-                .run_in_state(GameState::InGame)
-                .after("move_to_target"),
```
odecay, line 55: Alright I'm noticing this ordering has been lost. Is that accurate? There is a chance that its unnecessary if you are saying everything is working as expected, but if there was an issue it would be effecting the targeting systems.

src/main.rs
```diff
@@ -203,44 +314,7 @@ fn main() {
...
-        .add_system(
-            set_target_near_player
-                .run_in_state(GameState::InGame)
-                .label("set_target_near_player"),
-        )
-        .add_system(
-            move_to_target
-                .run_in_state(GameState::InGame)
-                .after("set_target_near_player")
-                .label("move_to_target"),
```
odecay, line 228: These are the other systems that were relying on the label ordering..

src/attack.rs
64kramsystem: So! This is a complex topic, and ultimately it depends on the design.

The ordering is still there, with these differences:

1. it's now implicit, because of the placement in the stages
2. it's now spread over two frames (previously, they'd run over the same frame, and did not suffer from 1-frame-lag, as events don't have this problem)
  - Frame 1: `Movement` stage -> `move_to_target`
  - Frame 2: `Action` stage -> `enemy_attack`

Now, we can place them in the same frame. This requires:

1. either moving the systems across stages (e.g. `move_to_target` to `Decisions`, however, this would be inconsistent)
2. or moving entire stages (e.g. `Movement` before `Actions`); this is a matter of just deciding the order of the stages, which is arbitrary to a certain extent

Right now, we don't have many options. [This](https://discord.com/channels/691052431525675048/1002614097050869780) is Alice's reply about my question on the topic, and this is the [Bevy stageless PR](https://github.com/bevyengine/rfcs/blob/2833d6bdb1386cef45c5617707be7947c25a3d68/rfcs/45-stageless.md).

Based on the reply and the PR:

- they're not going to implement automatic flushing based on ordering any time soon (I guess it's not even planned for now)
- the solution will be manual flushing via `chain![]` + `apply_buffers.named()` (see [here](https://github.com/bevyengine/rfcs/blob/2833d6bdb1386cef45c5617707be7947c25a3d68/rfcs/45-stageless.md#command-processing))

The big problem is that even with a theoretical manual flushing, once the number of systems reach a certain number, the dependency graph inevitably becomes too big to be manageable, so we'll need to split in stages, in order to make the ordering understandable. And once we split in stages, we're back to the model of this PR/Issue.

I can't think of other realistic, manageable design, aside:

- a fully serial one (of course, this kinda defeats the point of using Bevy)
- a, let's say, "semi-serial" one, with much smaller stages, which has _some_ level of parallelism, but is substantially serial.

It's certainly possible that I may miss other designs, so I'm open to any ideas :smile:. I think though, that we need to tackle the 1-frame-lag and ordering, in this stage of development, otherwise, once the number of systems increase, it's going to be insane™ :exploding_head:

odecay: > it's now spread over two frames (previously, they'd run over the same frame, and did not suffer from 1-frame-lag, as events don't have this problem)

The reason I designed it to use events was to avoid that one frame lag issue. This is the sort of thing I was advocating for continuing to keep while it is useful in  https://github.com/fishfight/punchy/issues/99#issuecomment-1192243114. Events and system order let us have dependencies between systems without either waiting for a frame for a command to apply or moving systems into arbitrary stages. I don't think we need to fit everything into stages design as our only tool to solve ordering constraints.

I have read and been keeping up with the stageless discussion and know that it will require manual flushing.
I think my preferred design would be to use a combination of the tools available to us, with stages in current design or manual command flushes after certain system sets in stageless, but also without entirely removing the possibility of using events and label based system ordering when we need them/when we want more granular control or when there are a few systems which depend on each other but which don't have much implication in the larger dependency graph.

I don't think its realistic to aim for a fully parallel design, and I am unsure how we would accomplish that given the dependency and ordering tools we have available.
It's my understanding that this design with stages or the future design with manual flushing where stage boundaries currently are *is* semi serial. Its blocks of systems which must run one after another, but within the blocks (stages) systems are allowed to run in parallel.

64kramsystem: > The reason I designed it to use events was to avoid that one frame lag issue

Ok, I can change the PR. Which is your concrete proposal? I can think of the two options, mentioned above (moving the systems across stages or moving entire stages), but there may be different approaches.

odecay:
Hmm my previous comment was showing up twice so I deleted one of them but now both are gone, thanks github.

64kramsystem:
> Hmm my previous comment was showing up twice so I deleted one of them but now both are gone, thanks github.

Ouch :grin: Didn't see it on time :flushed: I've checked the inbox, but it seems that GitHub did not send me a notification email.

64kramsystem:
Closing this as well - it started to drift, and it would also require to account systems introduced in the meantime.

Also, to clarify the rationale: this is also orthogonal to state mangement; the rationale is primarily to manage commands flushing, which is currently a problem.

zicklag:
This might still end up being something we want to do. I'm not sure yet.

Honestly I've stayed out of the gameplay aspect of this game until I got into thinking about states so I've got to do a lot of interacting with the code before I get an accurate picture of where we are and how things are working.

So I hadn't been able to give any accurate review on these things, and then I had the idea I wanted to try out for states, so I thought working on that would help me understand the problem space.

I don't want you to feel like I'm trying to come in with some "big plan" that just shuts down your work or anything. My idea for states just kind of touches lots of different things and I want to know what impact it will have on scheduling and stuff.

For instance, I found a way to create custom command buffers that will flush before the end of the frame, which could help solve some of our problems, but your stages proposal could still be something we want to implement to provide more determinism on system ordering.

64kramsystem:
> This might still end up being something we want to do. I'm not sure yet.
> 
> I don't want you to feel like I'm trying to come in with some "big plan" that just shuts down your work or anything. My idea for states just kind of touches lots of different things and I want to know what impact it will have on scheduling and stuff.

Ook, no poblem :smile: The important thing is that the flushing design is handled in one way or another, since the project does currently suffer from 1-frame-lag bugs.

PR #186, opened 2022-07-31
Implement trip points

### LINKED ISSUE #183: Implement trip points
### Description

A short time ago, I've implemented a trivial FOV (for the enemies). Its main purpose was to avoid all the enemies on the level to run immediately towards the player.

The next step is tripping points. Each enemy on the level will be assigned a tripping point, which will allow the enemy to move only after any player trips.

This actually allows cool things. For example, in Final Fight, there were enemies standing on the wall, waiting behind barrels (and then kicking them) etc.etc.

I'm scheduling this for 0.0.4, because I should have capacity.

### Alternatives & Prior Art

_No response_

64kramsystem:
Simplest implementation of trip points (replaces enemy FOV).

A trip point is associated to each enemy, which don't set a target unless any player ever trespassed the point. Once a point has been trespassed, the given trip point is set to f32::MIN.

This can be easily changed in the future, for more sophisticated logic (e.g. enemy state change).

Closes #183.

### DIFF
--- a/src/metadata.rs
+++ b/src/metadata.rs
@@ -135,7 +135,12 @@ pub struct FighterSpawnMeta {
     pub fighter_handle: Handle<FighterMeta>,
     pub location: Vec3,
     // Set only for enemies.
-    pub trip_point_x: Option<f32>,
+    #[serde(default = "default_f32_min")]
+    pub trip_point_x: f32,
+}
+
+fn default_f32_min() -> f32 {
+    f32::MIN
 }
 
 #[derive(HasLoadProgress, TypeUuid, Deserialize, Clone, Debug)]
--- a/src/metadata.rs
+++ b/src/metadata.rs
@@ -134,6 +134,8 @@ pub struct FighterSpawnMeta {
     #[serde(skip)]
     pub fighter_handle: Handle<FighterMeta>,
     pub location: Vec3,
+    // Set only for enemies.
+    pub trip_point_x: Option<f32>,
 }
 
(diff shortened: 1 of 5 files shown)


src/metadata.rs
```diff
@@ -134,6 +134,8 @@ pub struct FighterSpawnMeta {
     #[serde(skip)]
     pub fighter_handle: Handle<FighterMeta>,
     pub location: Vec3,
+    // Set only for enemies.
+    pub trip_point_x: Option<f32>,
 }
```
zicklag, line 139: Do you think it makes sense just to have a default value here so we don't have to worry about unwrapping it later?

Threw a doc comment in there, too, just 'cause I had the time. :)

> **Note:** Suggestion is pseudo code and not tested.

```suggestion
    /// The point in the level players must pass for this enemy to start moving for the first time.
    ///
    /// This is only meaningful for enemy spawns, not for player spawns.
    ///
    /// A value of [`f32::MIN`] indicates that the enemy should start starts off moving by default.
    #[serde(default = default_f32_min)]
    pub trip_point_x: f32,
}

fn default_f32_min() -> f32 {
    f32::MIN
}
```

64kramsystem: It changes the semantics, however, on the other hands, it's a simplification without unwanted side effects, so I'll change it (requires also the unwrap() removal) :+1:

64kramsystem: Updated :smile:

Issue #144, opened 2022-07-25
Remove `debug` Feature In Favor of Runtime Configuration


odecay:
### Description

Would it make sense to include a `cargo clippy --features=debug` somewhere in our CI?
I have noticed on multiple occasions that debug features end up broken since we currently have no checks for them, and when refactoring it is easy to forget to unregister a component from bevy-inspector-egui or some other small thing. If it takes too much time and needlessly slows down our workflow for commits I don't mind not having it, since debug features are not really user facing and I have currently been dealing with it by just doing the small fixes required.

There is also `--features=schedule_graph` but this is used even more rarely so I'd probably advocate for not checking against it.

### Alternatives

Not doing this if it causes annoyance to PR workflow.

zicklag:
I wonder if maybe we could take another approach and remove the feature completely, allowing the `debug` features to be enabled at runtime with a engine config flag.

That would prevent the different codepaths and CI checking, and it would allow more powerful debugging features for users who decide to test out modding without having them need to download a debug enabled build or else compile it themselves.

64kramsystem:
> I wonder if maybe we could take another approach and remove the feature completely, allowing the debug features to be enabled at runtime with a engine config flag.

I think that's a good idea. I always hate greyed out code :joy:

odecay:
I wouldnt mind that, it would remove the annoyance of not having rust-analyzer pick up the debug features as well.

odecay:
I dont know if that would work for the schedule graph? But that also is much less essential to have checked than debug features, and much less likely to break.

zicklag:
> I dont know if that would work for the schedule graph? 

I feel like we could leave that one as a feature flag. It seems like one that's rare even for developers and would probably never be used by modders.

odecay:
Yeap sounds good to me.

PR #189, opened 2022-08-01
Update to Bevy 0.8 🎉

zicklag:
It works on Bevy 0.8!

Currently we're depending on the Bevy default features, but that pulls a lot more than we need. I'm going to follow up with a PR to try and cut that down. We have some dependencies depending on Bevy UI, PBR, and Audio, I think and I need to figure out which ones.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -91,6 +91,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "android_system_properties"
+version = "0.1.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d7ed72e1635e121ca3e79420540282af22da58be50de153d36f81ddc6b83aa9e"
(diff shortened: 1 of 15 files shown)


64kramsystem (approved):

odecay (commented):
Looks good to me :+1:
we will be waiting on the debug stuff to return, looks like `--features=debug` wont work till then

zicklag:
Yep, we can follow up with that as soon as those crates update, and we might as well tackle #144 at the same time.

odecay:
> Yep, we can follow up with that as soon as those crates update, and we might as well tackle #144 at the same time.

sounds good

odecay:
[pr for update on bevy-egui-inspector](https://github.com/jakobhellermann/bevy-inspector-egui/pull/76)

PR #190, opened 2022-08-01
Update YAML Version and Cargo Update

zicklag:
This updates the last dep we had out of date, `serde_yaml`, which uses a new syntax for enum tags:

**Before:**

```yaml
        up:
          SingleAxis:
            axis_type:
              Gamepad: LeftStickY
            positive_low: 0.1
            negative_low: -1.0
```

**After:**

```yaml
        up: !SingleAxis
          axis_type: !Gamepad LeftStickY
          positive_low: 0.1
          negative_low: -1.0
```

It's a little different than I'm used to. I didn't know YAML had tags, I think it might be a YAML 2.0 feature or something, but I what I _do_ like is that it fixes the inconsistency of using `snake_case` for all of the keys ( leafwing input didn't rename the fields to use snake case like we did ). Since enum tags are now using the `!` syntax and that feels a little bit more consistent and valid to have them use the CamelCase for the tags.

---

Another note is that I had to add a custom deserialize implementation for keeping our `[start_frame, end_frame]` syntax for animation clip ranges. Serde deserialize implementations are a little verbose, but it didn't turn out that bad.

### DIFF
--- a/assets/default.game.yaml
+++ b/assets/default.game.yaml
@@ -21,72 +21,50 @@ default_settings:
     # Gamepad controls
     gamepad:
       movement:
-        up:
-          SingleAxis:
-            axis_type:
-              Gamepad: LeftStickY
-            positive_low: 0.1
-            negative_low: -1.0
-        left:
-          SingleAxis:
-            axis_type:
-              Gamepad: LeftStickX
-            positive_low: 1.0
-            negative_low: -0.1
-        down:
-          SingleAxis:
-            axis_type:
-              Gamepad: LeftStickY
-            positive_low: 1.0
-            negative_low: -0.1
-        right:
-          SingleAxis:
-            axis_type:
-              Gamepad: LeftStickX
-            positive_low: 0.1
-            negative_low: -1.0
-      flop_attack:
-        GamepadButton: South
-      shoot:
-        GamepadButton: East
-      throw:
-        GamepadButton: West
+        up: !SingleAxis
+          axis_type: !Gamepad LeftStickY
+          positive_low: 0.1
+          negative_low: -1.0
+
+        left: !SingleAxis
+          axis_type: !Gamepad LeftStickX
+          positive_low: 1.0
+          negative_low: -0.1
+
+        down: !SingleAxis
+          axis_type: !Gamepad LeftStickY
+          positive_low: 1.0
+          negative_low: -0.1
+
+        right: !SingleAxis
+          axis_type: !Gamepad LeftStickX
+          positive_low: 0.1
+          negative_low: -1.0
+      flop_attack: !GamepadButton South
+      shoot: !GamepadButton East
+      throw: !GamepadButton West
 
     # Controls for the first keyboard player ( left side )
     keyboard1:
       movement:
-        up:
-          Keyboard: W
-        down:
-          Keyboard: S
-        left:
-          Keyboard: A
-        right:
-          Keyboard: D
-      flop_attack:
-        Keyboard: Space
-      shoot:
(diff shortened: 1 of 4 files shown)


assets/default.game.yaml
```diff
@@ -21,72 +21,50 @@ default_settings:
...
+          positive_low: 1.0
+          negative_low: -0.1
+
+        right: !SingleAxis
+          axis_type: !Gamepad LeftStickX
+          positive_low: 0.1
+          negative_low: -1.0
+      flop_attack: !GamepadButton South
+      shoot: !GamepadButton East
+      throw: !GamepadButton West
```
odecay, line 45: It's also less lines which is nice :shrug:

src/animation.rs
```diff
@@ -45,11 +46,53 @@ impl Facing {
...
+            return Err(serde::de::Error::invalid_length(
+                1,
+                &"a sequence with a length of 2",
+            ));
+        };
+
+        Ok(start..end)
+    }
+}
+
```
odecay, line 95: This reminds me of something I was wondering about before, which is not entirely relevant here, but I wonder if we can implement clips/animations which play a non-contiguous array of frame indexes. I had a usecase for it before when I was adding a waiting state and it didnt really match any of the available animation frames, but I'm not sure if that would be a good design. It would let us piece together frames from a spritesheet to make "ad-hoc" animations of the correct length, or skip frames in the middle of an animation if we needed to. The more I write it out the more janky it sounds though so maybe this is not a good idea :laughing:

assets/default.game.yaml
zicklag: Yeah, overall seems nice.

src/animation.rs
zicklag: It could make sense. It would essentially be an array of ranges, which might not be bad.

Some of the exact details of the syntax become less important with an editor, too, so if it was a useful feature and we could make a good interface for it, it seems like it might be good to include.

zicklag:
We good to merge this then?

odecay:
@zicklag this broke level loading, getting past Main Menu panics
`thread 'main' panicked at 'Get value from storage: SerializationError(Error("invalid type: map, expected a Value::Tagged enum"))', src/loading.rs:358:25`

zicklag:
Ah, I forgot about that.

The issue is that the format for the local storage file changed with the new YAML format. You just have to delete the storage YAML file and start the game again and you'll be good.

On linux the file is `~/.local/share/punchy/storage.yml`.

This does bring up something we need to figure out. How do we gracefully fail when the YAML format changes or is otherwise corrupted and we can't read it?

I think the best option is to move the old storage file to `storage.yml.bkup` and then just re-create it with default settings. I'll open a PR to do that. On WASM, I suppose we just clear the user's data and replace it with default?

I don't really like that behavior in general, but at least while the game is unstable I feel like it makes sense.

odecay:
Ohh interesting, so then it wasn't *really* broken? I didn't think of the locally stored settings causing the issue.
What do you want to do with this now while we figure out what to do about local settings storage/format change since I reverted in https://github.com/fishfight/punchy/pull/193?

zicklag:
I'll just open another PR that un-reverts it, and then add another commit to provide a non-panicky fallback on failure to deserialize the save data.

I'm thinking the simplest solution is best for now and I'll just delete the save data if it fails to deserialize. We can re-visit when we have save data that's important, but I don't think we really have to worry about a user losing their hard-earned progress right now because there's no progress to earn. :grin: Even if there was, they could just edit the YAML of their own save data.

PR #192, opened 2022-08-01
Remove Unneeded Dependencies

zicklag:
Brings our dependency count for builds ( on linux ) from 404 to 376 ( 28 dependencies removed! ). Once https://github.com/dimforge/bevy_rapier/pull/225 is merged we can save an additional dependency on `bevy_pbr`.

Some of the more suspicious dependencies I had noticed that this removes are `bevy_ui` and `bevy_gltf`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ members = [
 
 [dependencies.bevy]
 version = "0.8.0"
-default-features = true
+default-features = false
 features = [
     "x11",
     "png",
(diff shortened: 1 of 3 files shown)


zicklag:
Merging because this is _super_ simple. :)

bors r+

Cargo.toml
```diff
@@ -14,7 +14,7 @@ members = [
 
 [dependencies.bevy]
 version = "0.8.0"
-default-features = true
+default-features = false
```
odecay, line 17: do we need to add back any of the features here? https://github.com/bevyengine/bevy/blob/main/docs/cargo_features.md looks like we would be missing winit and render other important looking things? (good chance I'm misunderstanding)

odecay:
Idk seems like something broke

Cargo.toml
zicklag: Ah, I'll add those to be explicit, but since our dependencies depend on those we didn't need to specify.

We should put those in there, though, I'll follow up.

odecay: Yea it seems like it didn't cause any issues so if its unnecessary thats okay.

PR #194, opened 2022-08-01
Deserialization Fix

zicklag:
Un reverts https://github.com/fishfight/punchy/commit/a60ed4a4864f1c7ff8f5a0633840c94cd9676876 and adds a more graceful fallback to deserialization problems.

The fallback right now is to log an error when we fail to deserialize something, but we just act like the data isn't there afterwards. This means that subsequent attempts to save will overwrite the non-understood value.

This way we get some notification that something unexpected happened, but we don't panic and we can gracefully continue without having to manually modify files on the filesystem or browser storage.

### DIFF
--- a/src/platform.rs
+++ b/src/platform.rs
@@ -169,7 +169,16 @@ impl Storage {
     where
         T: Serialize + DeserializeOwned,
     {
-        self.try_get(key).expect("Get value from storage")
+        match self.try_get(key) {
+            Ok(data) => data,
+            Err(e) => {
+                error!(
+                    "Error deserializing storage key, returning empty value: {}",
+                    e
+                );
+                None
+            }
+        }
     }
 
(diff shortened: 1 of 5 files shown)


src/platform.rs
```diff
@@ -383,8 +406,17 @@ mod wasm {
...
                             .unwrap()
-                            .and_then(|data| {
-                                serde_yaml::from_str(&data).expect("Deserialize platform storage")
+                            .and_then(|data| match serde_yaml::from_str(&data) {
+                                Ok(data) => data,
+                                Err(e) => {
+                                    error!(
+                                        "Error deserializing storage, ignoring existing data \
+                                        and overwriting on next attempt to set data in storage: {}",
+                                        e
```
odecay, line 415: :+1:

PR #195, opened 2022-08-01
Implement intention to move

64kramsystem:


64kramsystem:
@zicklag @odecay This is a massive pain in the back, and I've spent considerable time without being able to make it work.

This PR is intended to implemented the intention to move, independently of the work on the stages, by using before/after associations. I think this problem is not solved by stages, so this implementation can stay independent.

Now, something unexpected happens - the systems now step onto each other, likely because the associations have changed the way Bevy schedules them.

## Summary

Premise: on each stage, the systems send intention to move, conditional to their state; a flop attack can block player controller events, by setting the state to attack.

Now, let's imagine a linearized history:

- the user taps attack
- the player controller system runs first, and seeing that the state is idle, sends an intention event
- the player flop system then runs, and changes the state to attack (but waits for the next frame to start the actual attack)
- the events processor reads the controller event, which sets the state to idle, nullifying the flop attack

Under regular circumstances, something like this would happen

- the user taps attack
- the player controller runs first, sees that the state is idle, sets the state still as idle
- the player flop system runs, and changes the state to attack (but waits for the next frame to start the actual attack)
- next frame: the player controller runs first, but the state is attack, so it doesn't take any action
- the player flop system runs, find that the state is attack, and proceeds with the attack.

Amusingly, in some game runs, the flop runs fine, which I guess happens because Bevy schedules the systems in an inverse order.

I think that this problem can't be solved without centralizing the player state machine into a single system, because without such change, state-related systems are going to step onto each other (more or less, depending on them being split or not).

Note that currently, we may have some very subtle misbehaviors, but they're not evident because they're unlikely. For example:

- the player controller system runs
- the player flop system runs concurrently
- the player flop system finishes first, and sets the state to attack
- the player controller system finishes second, and overwrites the state to running.

I can imagine that if Bevy will make changes to the scheduler across versions, without a strict design, we're going to be in a _lot_ of pain :grimacing:

I'm not sure how I should proceed here.

zicklag:
I may understand wrong, so whatever I say might not make sense, but here's some thoughts:

It sounds like our issue is essentially the arguable number one cause of bugs in programs: global mutable state.

It seems like the solution then is to find a way to make the state non-shared across systems, as much as appropriate. In other words, we probably shouldn't have multiple systems that write to the player state. But we could have multiple systems that send messages about what they _want_ to attempt to write to state, maybe.

What if we did something like this:

- All systems that collect user intentions run in any order relative to each-other and they send events indicating their intent.
  - For instance, the player controller system sends intents to move
  - The flop system sends intents to flop
- After all those systems run, there is another system that reads all the events sent by those systems
  - This system will process all the intents and use whatever logic makes sense to filter out the intents that should not be acted on.
    - For instance, it will not move the player if we are in the middle of flopping
  - This system will, based on the non-filtered events, apply changes to the player state.
  - It's the only system that modifies the player state based on intents, so we don't end up with multiple systems all trying to modify the state of the player at the same time or in a random order.

I haven't done a lot of looking into what our code is now or what your guys' proposals were earlier so this may or may not make much sense.

odecay:
Okay there is a lot here,
I will try to read through and see what parts I can address.

> I can imagine that if Bevy will make changes to the scheduler across versions, without a strict design, we're going to be in a lot of pain

Bevy always schedules systems non-deterministically so if they are not explicitly ordered this will happen. So I think you are correct here.

I was under the impression that we would be emitting events with intent to move in every system that currently directly moves the player.

> the events processor reads the controller event, which sets the state to idle, nullifying the flop attack

There was a similar problem awhile back, where sometimes state would get reset before playing the animation. I solved this by moving the return to idle function to its own stage at the end of frame. In this case I think the event processor should live in a stage toward the end of the frame, and be reading not the controller events but the "intent to move" events, and deciding which of them are valid based on the players state when its stage is reached (within the same frame).

>The player flop system then runs, and changes the state to attack (but waits for the next frame to start the actual attack)

Are you referring to the fact that the attack is spawned via commands? State is set immediately to `State::Attacking` in `player_flop` but the commands dont technically have to wait till the next frame to be applied, since they apply on stage transition.

 Apologies for a fragmented response. I will try to review the code that is here and see what I can make of it, these were just some things that stood out to me right away.

64kramsystem:
> I was under the impression that we would be emitting events with intent to move in every system that currently directly moves the player.

If I understand correctly, I've done this in the PR, emitting intent to move events in three cases:

- player controller response
- knockback
- flop

>> The player flop system then runs, and changes the state to attack (but waits for the next frame to start the actual attack)

> Are you referring to the fact that the attack is spawned via commands? State is set immediately to `State::Attacking` in `player_flop` but the commands dont technically have to wait till the next frame to be applied, since they apply on stage transition.

Sorry, I've noticed I've used the wrong term - wtih "attack" I was referring, more precisely, to the movement.

In the logic [here](https://github.com/64kramsystem/punchy-dev/blob/master/src/attack.rs#L310-L353) - when the flop is invoked, it it sets the [attack state](https://github.com/64kramsystem/punchy-dev/blob/master/src/attack.rs#L314), and prepares the collider etc. but doesn't trigger the movement. This is correct (it's probably not correct to move immediately).

This was just an observation, anyway. The problem lies in the fact that by splitting into sender/receiver systems, there are scheduling problems.

>> the events processor reads the controller event, which sets the state to idle, nullifying the flop attack

> There was a similar problem awhile back, where sometimes state would get reset before playing the animation. I solved this by moving the return to idle function to its own stage at the end of frame. In this case I think the event processor should live in a stage toward the end of the frame, and be reading not the controller events but the "intent to move" events, and deciding which of them are valid based on the players state when its stage is reached (within the same frame).

I'll need to think this very carefully. It is definitely solvable, I was wondering what's a clean and idiomatic design. I'm putting my ideas in a separate comment.

64kramsystem:
So, based on responses, throughts, etc., the (IMO) cleanest and idiomatic design is the following.

## High level overview

- a single system processes the player state (currently, at least the two systems player_controller and player_flop do)
- it dispatches a signal
- systems that process state (controller, flop, knockback...) run conditionally to the signal

## Implementation details

The nature of the signal is a problem with Bevy/iyes_loopless.

1. If we use events, all the systems will be executed, and will need to have each a conditional in the function body. This works, but it's definitely unidiomatic in terms of ECS design, but also sofware design in general
1. Resources work as well, however, while on one side they're idiomatic (iyes_loopless supports conditional system execution using resources), this mean that each player state should be moved to a resource, which is unidiomatic (in terms of ECS design)
1. Components work, and they're idiomatic, but then they make multistage design a requirement - it's not a nice-to-have thing anymore, since the components will need to be flushed.

Option 3 is the only one I can think of (I didn't put a lot of thought, though). I've checked the ordering dependencies, and we don't need to go full-blown multistage. Right now, I think most of the logical sequences run with a 1 frame delay, so we don't need an exact sequence.

To be clear about the 1 frame delay, I don't take it as a problem - I do think it's a problem only when it's not deterministic and/or not planned.

64kramsystem:
Made some edits to the last comment.

64kramsystem:
(other options I can think of; cross-posting from [Discord](https://discord.com/channels/691052431525675048/746807925107851346/1003985437771169812))

Some other options I can think about, by looking at the `iyes_loopless` APIs:

5. it seems that full-blown systems (that is, using Query) can be used as conditions; this is _really_ neat if used with const generics, but unfortunately we're on stable, and without enum support, we'd have to create one system for each condition. just to be clear, this is what I'm talking about:
    ```rs
    // requires one per state
    fn is_player_idle(q: Query<&State, With<Player>>) -> bool {
        matches!(q.single(), State::Idle)
    }
    // const generic version - only one definition; the compiler generates one per state
    fn is_player_in_state<const S: State>(q: Query<&State, With<Player>>) -> bool {
        matches!(q.single(), S)
    }
    ```
6. event-based conditionals (`run_on_event`) are also supported; by creating a corresponding event type for each state, an API `State#send_event(w: EventWriter)` could be created (with a matcher inside)

Of all the solutions, the only satisfactory one I can think of is option 6. However, I certainly may be missing something, and regardless, I'd like some input on the topic.

zicklag:
> If we use events, all the systems will be executed, and will need to have each a conditional in the function body. This works, but it's definitely unidiomatic in terms of ECS design, but also sofware design in general

This doesn't feel too bad to me, personally. At least the intent is clear and you can even reduce rightward pressure by returning early from the system if the event you are interested in hasn't been triggered.

> event-based conditionals (run_on_event) are also supported; by creating a corresponding event type for each state, an API State#send_event(w: EventWriter) could be created (with a matcher inside)

This also seems like a nifty use of conditions and a good option.

zicklag:
OK, I had some ideas, but it got long enough I put it in a discussion instead: https://github.com/fishfight/punchy/discussions/196.

odecay:
I'm going to try to dig into @zicklag proposal when I get the chance but I also think that 1, 3, 5, 6 could all be acceptable. I think we will be unlikely to have an unmanageable amount of states so I don't think matching will be too much of a hassle, although I do remember thinking it would be nice to figure out a unified method for state transition/allowing or disallowing states to be entered based on current state.

64kramsystem:
To clarify, the intention to move design is orthogonal to state management.

The rationale is to decouple systems, specifically, to decouple the clamping from the movement origins; right now, the clamper needs to be invoked in any system where player movement is involved.

However, this is not something strictly necessary. I'm going to close this, since it's starting to drift, and it's unlinkely to be picked up.

64kramsystem:
Just for the lulz, somebody on discord gave me a really neat solution to the conditional system run design:

```rs
fn player_action() { /* ... */ }

fn in_state(target_state: PlayerState) -> impl Fn(Query<&PlayerState, With<Player>>) -> bool {
    move |q| *q.single() == target_state
}

app.add_system(player_action.run_if(in_state(PlayerState::State1)))
```

This may turn useful in the future.

zicklag:
> I'm going to try to dig into @zicklag proposal when I get the chance

Hopefully I'll have some concrete code to go through soon. I've just about got the player actually idling and flopping with my proposed design and I want to see how far I can being it in a reasonable time frame so we can get a real idea of what it will look like.

> In another game I've used components to represent state and found it works relatively well aside from having to wait for commands 

I found a way to create a custom command buffer that's gets flushed before the normal commands, before the state handler systems actually run. So now I've got the following "stages" in my prototype ( which may or may not end up literal bevy stages ):

- Collect transition intents
  - This is where all the systems that nay express t intent to transition the player state will go.
- Compute state transitions 
  - This contains the state transition logic ( this is a new concept ).
  - For each state there will be a `transition_from_[state]` system that will read intents to transition the state and determine whether there are indeed any state transitions.
  - These are the systems that will use our command buffer to add and remove components
- Flush state transitions
  - This is a system that flushes the state transition command buffer, immediately applying any component updates requested by the "Compute state transitions" stage.
- Handle states
  - This is where the real state systems are, the ones that modify the player based on their current state, such as playing animations, etc.

> I'm closing this PR, since it's starting to drift, and it's unlinkely to be picked up.

Thanks for the work either way! I know how disappointing it can feel to drop a PR you've spent a lot of time on.

> To clarify, the intention to move design is orthogonal to state management.

Yeah, the original point of this PR was not supposed to include state management, but then the issues around scheduling kind of bled into how we are going to schedule the handling of the player state in a way that makes sense and doesn't produce so much bugs.

Hopefully once we fix that the intent to move issue will be an easy fix.

PR #200, opened 2022-08-02
Add boss MVP

edgarssilva:
The beach level only has the boss for easier debugging. I usually use the bandit sprite since I'm waiting on the boss ones (I might give it a shot myself) but currently it's using the current boss sheet (very distracting).

The temporary way to define the boss is by passing boss: true in yaml file of the level, this will get picked up and add a Boss marker component to the enemy entity.
This will make the enemy not use the default attack and instead do a jump one though sometimes doesn't land on the current spot. Numbers should be adjusted but waiting for the textures to fit it nicely.


64kramsystem:
In general, looks good to me, as it just follows the same patterns we use. The new `boss` is necessary, as there is special boss behavior (e.g. when the boss dies, all other minions die as well, level is completed, etc.).

Something that I can imagine in the future (near or far :smile:) is for the attacks to be defined by metadata. However, I think this is something quite complex, since it will also interact with scripting, states etc, which are under heavy work.

In summary, LGTM :smile:

64kramsystem (commented):
In general, looks good to me, as it just follows the same patterns we use. The new `boss` is necessary, as there is special boss behavior (e.g. when the boss dies, all other minions die as well, level is completed, etc.).

Something that I can imagine in the future (near or far :smile:) is for the attacks to be defined by metadata. However, I think this is something quite complex, since it will also interact with scripting, states etc, which are under heavy work.

In summary, LGTM :smile: 

Needs to be rebased.

64kramsystem:
Ah, BTW, I've marked the PR as draft.

edgarssilva:
Just got the boss assets, @zicklag not sure what to do with attack conflicts. Could you take a look?

zicklag:
Sure thing, I'll check it out as soon as I get the chance.

src/attack.rs
```diff
@@ -494,9 +490,9 @@ fn boss_attack(
                         .insert(Attack { damage: 30 })
                         .insert(AttackFrames {
                             //TODO: Check if this is correct
-                            startup: 1,
-                            active: 2,
-                            recovery: 3,
+                            startup: 6,
+                            active: 4,
+                            recovery: 5,
```
odecay, line 495: startup, active and recovery are all measured from the beginning of the animation.
I was thinking about if there was a better way to represent this than ints, like ranges of frames, to better communicate intent here.

The startup frame listed there is the last frame of startup, which starts at frame 0 of the animation and goes to `startup`.
The active frame is the last active frame, also measured from frame 0 of the animation.

(recovery is measured from beginning as well, right now its a bit redundant but the idea is that when we have chaining system in place it will be important to determine cancel windows for moves. It should not be > than the total frames of the animation, and conceptually represents the frames from when the hitbox deactivates, to the frame when you are able to cancel an attack/animation with a new input.) For now you can think of recovery as the last frame off the animation.

Right now as these are represented, their should not be a situation where startup > active, or active > recovery.

edgarssilva: So it's the frame index and not the number of frames? So 5 - 9 and 14 respectively.

odecay: Yup they are indexes

zicklag:
I found it easier to just copy the code from the diff on GitHub onto the latest master with the tweaks needed to make it work with the new fighter state machine.

I pushed to a branch in this repo that you can also push to and opened a new PR: https://github.com/fishfolks/punchy/pull/222. You can continue working on that branch.

edgarssilva:
Merged in #221.

Issue #17, opened 2022-06-28
Sort 2D sprites in 3D space


erlend-sh:
http://andrewrussell.net/2016/06/how-2-5d-sorting-works-in-river-city-ransom-underground/

![](http://andrewrussell.net/content/2016/06/Sort-Arrows.png)

https://www.youtube.com/playlist?list=PLgPMWQb41awm62M5YE3zzAKyw5Odka6JR

https://github.com/conatuscreative

zicklag:
@odecay @64kramsystem @edgarssilva I wanted to open discussion on this topic now because I'm wondering how far we want to go before we figure this out. It seems like it might be pretty important to do early.

Things like throwing items don't really make sense with our current 2D collision system, because if you throw the bottle in an arch it can go and hit fighters that are higher on the screen, even though it should technically be going up into the sky.

We'll need to position everything in 3D, and I think we will need to have 3D collision boxes for everything, including the players. The render position of the sprites on the screen will then be determined by a combination of it's `y` axis position and it's `z` axis position. Then we'd use something like the strategy described in @erlend-sh's links above for doing z-sorting of the rendered sprites.

I think fighter collisions will probably be simple enough, because it's just a square, but what will be more complicated is scenery. For example, if we have a car, we have to specify a multi-part 3D collision box for it, so that we have one section for the top of the car and the hood, etc. so you can jump up onto different parts of it. Specifying collision shapes for scenery might be an important role of an editor, and otherwise, we'll just have to deal with hot-reload combined with YAML definitions and debug rendering of the collision boxes.

Since rapier's 3D bevy plugin uses assumes you're using a 3D camera, and we are going to be using a 2D camera, we may also have to make our own rapier debug rendering system so that we can visualize the collision boxes in the space of our 2D rendering. We can do that simply enough with Egui, using the `epaint` API to render the debug lines.

---

I'm feeling like we should do this somewhat soon, before we get too stuck coding into a way that we are just going to have to un-do later, and I'd be find taking this on if nobody else wants to, but probably after getting #140 done ( which shouldn't be way to long from now ).

I'm wanting to try to get the foundations of the game nailed down so we don't have as many more PRs like #203 that kind of modify everything while other people are busy working on other parts of the game that are impacted by the big PR. Or at least I'd want to get those over with as soon as possible so people can keep adding feature in parallel.

But totally feel free to tell me what to do or that we have other ideas. Just let me know!

erlend-sh:
[...]

PR #203, opened 2022-08-03
New Fighter State Design and Gameplay Refactor

zicklag:
OK I think this is finished!

This refactors almost everything related to the gameplay 😅. It does an enormous amount of re-organizing, and clean up in addition to implementing the new fighter state design I started to describe in #196. I think a lot of it turned out easier to understand and simpler to modify.

I'm hoping this helps make it easier to maintain and easier to continue to add features as we continue to develop.

I think the only "regression" is that enemies jump around just like the player does when you attack, instead of just standing and punching, which should probably be fixed.

I want to wait to add different enemy attacks until this gets review, and we can probably do that in a separate PR.

---

Honestly, I kind of got carried away and I'll leave it up to you guys whether or not we should merge this an whether or not we do it all at the same time, if we do. If it's better, I can break this into smaller pieces and merge it one piece at a time ( as much as possible ). Removing the `State` enum alone pretty much necessitates a change to everything, so it can only be broken down so far, but I did re-organize a lot of things, too, so I could break the re-organizations into separate PRs if necessary.

## Summary Of Changes

@odecay this is mostly to help summarize the changes I made to aid in reviewing.

### Fighter State

The fighter and enemy state management changed completely. `fighter_state.rs` contains the entire fighter state machine. It's absolutely possible to split this out to separate files later if it gets too big, but I think it's still pretty easy to read through right now.

It uses the flow of: Collect Transition Intents ( from user input, enemy AI, or things like collisions ) → Evaluate Apply State Transitions → Run State Handlers.

The enemy AI logic is placed separately in `enemy_ai.rs`.

### Movement Model

I removed almost all places where we manually set an item/player translation and replaced it with a modification to a `LinearVelocity` component.

The `movement.rs` file has all of the systems that modify entity transforms based on linear and angular velocities and forces.

By having all our movement controlled by modifying velocities instead of translation, we are able to centralize the constraints, which are also placed in `movement.rs`. We are able to arbitrarily define new constraints on any kind of entity, be it player, fighter, item, etc. that operate on the Velocity of the entity before the the transform is applied. So we no longer have the issue of clamping fighter positions all over the place!

I also removed the `move_in_arch` system in favor of using forces and velocities to create the arch. I think this is cleaner, but that's subjective, so we could add move in arch again after this PR if you wanted. Though I think we might want to wait to do any more work in that direction until we figure out how we'll handle our collisions/coordinates/y-movement: #17.

### Attack and Damage

I made the damage system more generic, adding a `Health` component, a `Damageable(bool)` component and `DamageEvent` in `damage.rs`. I changed the player stats to record a `max_health`, so that we always know how much health they can go up to, and their actual health is stored in the `Health` component. This will allow us to add health to all kinds of stuff, such as destructible scenery, that can be damaged by attacks, without having to specialize the attack system for what it is attacking.

The attack system doesn't change much, but is now generic over anything with `Health` and `Damageable` and attacks now have a `velocity` that can be used to figure out which direction and how hard an attack should push what was attacked, if applicable. This is used for knockback on the player, but would be ignored if a wall was attacked for instance.

Damage events are emitted when an attack lands, and this event is used by the knockback collector system to put the player into a knockback state when they get attacked.

### Projectiles & Lifetime

I moved the `ProjectileLifetime` component to a generic `Lifetime` component, so we can use it on anything that we want to despawn on a timer.

### Despawn

I removed the `DespawnMarker` component and the `despawn_entities` system, because it's effectually exactly the same as just calling `commands.entity(entity).despawn_recursive()` which isn't really less ergonomic ( IMO ) than `commands.entity(entity).insert(DespawnMarker)` and doesn't require an extra import of the `DespawnMarker` component.

### Organization

I moved almost all remaining systems that were in `main.rs` to modules that made sense for them, adding plugins as necessary to each module that had systems in it.

I removed the `y_sort` module to move it's system and component in with the camera, and lots of other small organizations.

I think its easier to browse the codebase and find the systems that effect different parts of the game now.

### Tweak To Item Handling

Instead of matching on the item name to determine behavior, I add a `kind` field to the item YAML definition that allows you to specify what the item does. For health items you can specify how much health it refills, for throwable items you can specify how much damage it does.

Eventually I think we will want a dynamic way to handle item kinds, similar to the player state, so that they can be scripted and we don't have to have a central list of what kinds of items there are in an Enum.

But it's a good temporary solution.

### Enemies Attack Like the Player

Right now enemies and the player share the same attack, though they have different animations.

This is trivial to fix, but I didn't want to work on any more things non organizational/architectural before this is merged so that can be done afterwards.

### That's it! :upside_down_face: 

Other than the massive list of changes above, the rest of the game stayed the same. Enemy AI logic is the same, the player clamping logic was adapted to a constraint so trip points and camera movement all works the same. The UI didn't change. Etc.


zicklag:
@odecay and @64kramsystem I'd like to get your review on this when you get the chance. Since so much has changed, you might find it easier to browse the repo after the changes instead of doing a side-by side diff. GitHub1s can make this easier to do in the browser: https://github1s.com/zicklag/punchy/blob/fighter-state-refactor/src. The most important file is `fighter_state.rs` which implements a revised version of my fighter state proposal.

Again, I know this is a lot of changes and feel free to tell me I've got to break this down into more reviewable chunks.

edgarssilva:
I'm really liking all the changes especially the machine state for the fighter, looks very ergonomic and much more structured.

edgarssilva (approved):
I'm really liking all the changes especially the machine state for the fighter, looks very ergonomic and much more structured.

64kramsystem:
There have been two complete PRs of mine, that have been open for some time :thinking:

zicklag:
> There are two complete PRs of mine, that have been open for some time.

Yeah, sorry about that. While I was busy with this one it conflicts with almost everything and my local sandbox was a mess while it was in progress so switching to other PRs was difficult.

I'll look at those now.  :+1:

64kramsystem:
I'll take some time to review this well. Immediate question: does this branch run very fast (I reckon around 2x as fast) on anybody's machine? It does on mine, it's a odd :grimacing:

64kramsystem:
Other issues:

- on flop, the player end higher than it should
- not sure if it's related to the previous comment, but there are some sort of teleportations while running :laughing: I'll try to post a video, it's a bit subtle

64kramsystem:
Video of the teletransportations here:

https://user-images.githubusercontent.com/1595356/183265610-466f33f6-bef2-407c-aed0-9d4e122d2329.mp4

Look for small stuttering; during each stutter, the player ends up farther than it should

zicklag:
> It does on mine, it's odd grimacing

Uh, yeah, :grin: I didn't get to balancing the movements back to normal. If I remember correctly I think we were applying speeds previously without multiplying by the frame delta time, so when I started doing that I had to change all the fighters speeds and didn't get to making them match up better with their old speeds.

I didn't want to spend too much time on that until I got the architecture approved, but this definitely needs another pass afterwards to bring the balancing back to normal.

> on flop, the player end higher than it should

That must be because I took out the y-correction code. I thought it was just a hack to make the player not end up out of bounds, which isn't a problem with the new constraint system, so I took it out. I can put it back in there to fix it.

> not sure if it's related to the previous comment, but there are some sort of teleportations while running laughing I'll try to post a video, it's a bit subtle

I'll try to reproduce.

zicklag:
> Video of the teletransportations here:

Oh, they are walking way faster on your machine. I must have done something wrong in the velocity system. My monitor has a 144hz refresh rate and they go about half as fast as that. I'll look into that.

64kramsystem:
> > Video of the teletransportations here:
> 
> Oh, they are walking way faster on your machine. I must have done something wrong in the velocity system. My monitor has a 144hz refresh rate and they go about half as fast as that. I'll look into that.

No sure if it gives any hint or not, but my screen is a 240 Hz running at 60 Hz :shrug:.

64kramsystem:
> That must be because I took out the y-correction code. I thought it was just a hack to make the player not end up out of bounds, which isn't a problem with the new constraint system, so I took it out. I can put it back in there to fix it.

The flop y-value problem is a 1-frame-delay bug - when I redesigned the stages in order to ensure that writes and reads were applied in the intended order, that bug went away (and the self correction hack was not necessary anymore).

Now, the problem is that if this PR is intended also to solve 1-frame-delay bugs - at least the player-related ones - then it's not fully working as intended. However, I don't know if this PR intends to do so :smile:

zicklag:
> The flop y-value problem is a 1-frame-delay bug - when I redesigned the stages in order to ensure that writes and reads were applied in the intended order, that bug went away (and the self correction hack was not necessary anymore).

Could you maybe describe what's wrong there a bit more for me. I never fully understood what the issue is. I definitely want to solve any of the 1-frame bugs if possible with this so we don't have to re-design later.

What's the intended order?

I believe that everything proceeds consistently in this order:

Input manager ( collects gamepad button state ) → Collect state transition intents → Apply any required state transitions → State transitions are flushed → The state handler functions are run ( any player velocity changes are applied ) → The player velocity is read and the transform change applied → The frame is rendered

Is that incorrect, or is it just not doing what I think it is doing?

---

Also I notice on my machine I'm not rising at all while flopping. After 7 consecutive flops my y height is still the same, at least to within the size of a game pixel.

---

> No sure if it gives any hint or not, but my screen is a 240 Hz running at 60 Hz shrug.

Wow, I'm not sure what's going on there I might have to think about that. We might need to debug print and compare delta times or something. I don't see anything obvious that I got wrong in the velocity system:

```rust
pub fn velocity_system(mut query: Query<(&mut Transform, &LinearVelocity)>, time: Res<Time>) {
    for (mut transform, dir) in &mut query.iter_mut() {
        transform.translation += dir.0.extend(0.) * time.delta_seconds();
    }
}
```

zicklag:
I just looked at the old code and we were doing essentially the same thing, so there shouldn't be a huge difference like that. :thinking: 

```rust
                // Apply speed
                dir = dir * stats.movement_speed * time.delta_seconds();
```

But that also means that it doesn't make sense why I had to set the fighter velocities to `17000` instead of `150`. If we were applying the delta time in the old code, then I should have been able to keep the same player speeds.

I've got to be missing something...

64kramsystem:
So, first I'll post a video, so we have a reference (also for me, because I sometimes doubt of it myself :joy:):

https://user-images.githubusercontent.com/1595356/183266413-6b51674e-efdc-4e67-8413-b27f5a22002f.mp4

The bug is non-deterministic (from a user perspective), so I can imagine it not reproducing on some setups. Maybe you can try a 60 Hz screen speed and see if it triggers.

Regarding the reply, I'll split in another comment.

zicklag:
Oh, interestingly I was already running at 59.95hz. I just have forgot to switch it back to 144hz after playing a wine game, so this whole time I'd been testing at 59.95hz. Still, switching between both refresh rates doesn't change anything about the game for me. :man_shrugging:

64kramsystem:
> Could you maybe describe what's wrong there a bit more for me. I never fully understood what the issue is. I definitely want to solve any of the 1-frame bugs if possible with this so we don't have to re-design later.

> Input manager ( collects gamepad button state ) → Collect state transition intents → Apply any required state transitions → State transitions are flushed → The state handler functions are run ( any player velocity changes are applied ) → The player velocity is read and the transform change applied → The frame is rendered

A couple of clarifications 😬

I didn't intend to imply that the order I've established in the stages redesign is the one and only; sorry if I gave this impression 🙇

I have actually no idea what's the exact cause of the problem 🙁 This is because when I grouped the systems and divided them into stages, I did look inside the systems only up to a certain extent - I was more interested in their high-level overview, and to give them a consistent ordering.

This may be hard to get to the bottom of; especially because it's hard to compare the two approaches. Some hints may come from understanding how the move interacts with the animation frames (e.g. frames skipped due to unexpected scheduling; I think in some cases, something was skipped, but my memory is fuzzy).

zicklag:
> I didn't intend to imply that the order I've established in the stages redesign is the one and only; sorry if I gave this impression bow

Yeah, no problem at all. :) I wasn't thinking that I was just thinking if it worked earlier, than how was it working.

> I have actually no idea what's the exact cause of the problem slightly_frowning_face This is because when I grouped the systems and divided them into stages, I did look inside the systems only up to a certain extent - I was more interested in their high-level overview, and to give them a consistent ordering.

Got it. I'll try to compare the scheduling I used to the one you used in the other PR.

---

As far as the movement speed, I figured it out I think. I was setting a velocity and multiplying the velocity by the delta time! The whole point of velocity is that you set the desired speed and the velocity system applies the delta time as required so you don't have to. Working on fixing that now...

zicklag:
OK, fighter speed should be fixed. If you can, try to see if the flopping issue still happens for you. I haven't been able to reproduce yet.

zicklag:
Hey I just figured out how to reproduce. I created a utility function that will add a random thread sleep when the `random_debug_slowdowns` feature is applied. Adding this to the velocity system lets me reproduce the random flop issue!

So funny that that worked, but we might want to keep that feature around for use in the future.

Now I've just got to fix it. :)

edgarssilva:
Heard that bevy 0.8 had some updates on the way the delta is calculated, might have something to do with it 🤷

zicklag:
> Heard that bevy 0.8 had some updates on the way the delta is calculated, might have something to do with it shrug

I think I fixed that issue. Now I'm working on the flop, and I'm pretty sure I know why that's busted too, now.

The issue is that we're timing the flop velocity changes based on the _animation frame_. But animation frames are not updated granularly enough to be an accurate timer for adjusting the flop velocity.

For instance, if a frame takes too long, it will only advance one animation frame, but the flop velocity should have changed more because the delta on the last frame was higher than normal. Essentially, we should never time movement based on animation frames or we risk that kind of inconsistent behavior.

At least that's my theory, I'm trying to fix it now and we'll see if that works.

edgarssilva:
> 

Yeah seems about right. When I first implemented the jump it was a "cheesy" way of getting it done.

zicklag:
> Yeah seems about right. When I first implemented the jump it was a "cheesy" way of getting it done.

It definitely wasn't a bad way to start, and it looked nice. After messing with it a bit, I'm feeling like the only solution worth the time right now is just to use our little hack to set the player y to whatever it started at.

Eventually we've probably got to calculate player movement in 3D and use collision detection with the floor to find out when to stop moving down so investing more time in this right now probably isn't worth it, considering I'm fairly confident it's not a system ordering problem at this point.

edgarssilva:
As you said another big refactor we need is changing physics from 2d to 3d not sure how the angles for that work.

zicklag:
OK, I fixed all issues I'm aware of. Let me know if you find any more, or the fix doesn't work for your machine!

64kramsystem:
There's another bug; the enemies are now jumping when they attack.

64kramsystem (approved):

64kramsystem:
So!

I'm approving this so it can go in, however, I'm actually approving because it's very large, and I can't give a meaningful review, at least, without taking a very considerable time to understand it.

It's difficult to review because there are many commits, and on the other hand, when looking it in one pass, there are several large blocks that have been moved around.
I've had a look at the structure, and it's relatively monolithic, so I guess that at this point splitting it would be a very considerable effort. 

I'll put my considerations about the design in a separate comment.

64kramsystem:
On-demand flushing is very interesting. I think something like this should be published :smile: at least to make devs aware of what needs to and can be done.

A doubt I have is that, since (I presume that) this design deals only with fighter states, in the future at least, the other states will need commands flushing handling. However, I guess that it's assumed that non-fighter-state-related systems will be a very small minority, therefore, easy to handle, so I think that works.

On the other hand, this is a quite custom approach to commands flushing, which I think it has some downsides:

- it will be less intuitive to understand for devs not familiar with it
- it may be made obsolete by new Bevy versions
- it may need update work on new Bevy versions (I don't know how stable is the internal interface used)

That's all. Based on my understanding of the last commits, the Y-rise issue stands, but anyway, I think that, at this point, it can be worked on it seperately, so that this PR goes in.

Generally speaking, I'm... underwhelmed, by the large amount of work required to manage storage sinchronization, even  for a relatively simple game (less that 5k lines, counted by cloc), when using a parallel+asynchronous ECS (ie. Bevy). This is not related to anything in our game - it's my general impression of game engines design.

zicklag:
> I'm actually approving because it's very large, and I can't give a meaningful review, at least, without taking a very considerable time to understand it.

Yeah, I'm realizing it would have been good for me to give a summary walkthrough to help reviewing. I'll do that in another comment below. You don't have to go back and look at anything, but @odecay might find it helpful.

> There's another bug; the enemies are now jumping when they attack.

That's a simple fix that we can do after merge. I wanted to avoid adding more gameplay code that could easily come after, so I just shared the attack system for the enemies with the players.

> On-demand flushing is very interesting. I think something like this should be published smile at least to make devs aware of what needs to and can be done.

I know I was thinking that would probably make a good crate. I think the idea is somewhat similar to the new command flushing design in the [Bevy Stageless RFC](https://github.com/alice-i-cecile/rfcs/blob/stageless/rfcs/45-stageless.md). 

Since `iyes_loopless` is kind of a preview for stageless I wonder if they'd like the feature in their crate.

> in the future at least, the other states will need commands flushing handling.

That's why `CustomCommands` is generic over a marker trait. We can add as many custom command queues as we want that get flushed whenever we want based on where we add the `flush_custom_commands::<MyCustomCmds>` system to the system schedule. So if other parts of the game need command flushing, we can just add it wherever we need it.

> which I think it has some downsides

The implementation is super simple, and the only change to the systems themselves is that they call `let commands = custom_commands.commands();` and use `CustomCommands<Marker>`, so I think it'd be really easy to migrate away if we needed to.

And if the Bevy Ttageless proposal gets accepted, we can just migrate to that, as it provides the same functionality.

So if we needed to update it would probably take me less than half-an-hour to do it.

> Based on my understanding of the last commits, the Y-rise issue stands, 

Yep, I'm pretty sure that's a "physics" issue, because we need to decide how to handle our "up" axis, not a scheduling issue, so it shouldn't go into this PR.

zicklag:
## Summary Of Changes

@odecay this is mostly to help summarize the changes I made to aid in reviewing.

### Fighter State

The fighter and enemy state management changed completely. `fighter_state.rs` contains the entire fighter state machine. It's absolutely possible to split this out to separate files later if it gets too big, but I think it's still pretty easy to read through right now.

It uses the flow of: Collect Transition Intents ( from user input, enemy AI, or things like collisions ) → Evaluate Apply State Transitions → Run State Handlers.

The enemy AI logic is placed separately in `enemy_ai.rs`.

### Movement Model

I removed almost all places where we manually set an item/player translation and replaced it with a modification to a `LinearVelocity` component.

The `movement.rs` file has all of the systems that modify entity transforms based on linear and angular velocities and forces.

By having all our movement controlled by modifying velocities instead of translation, we are able to centralize the constraints, which are also placed in `movement.rs`. We are able to arbitrarily define new constraints on any kind of entity, be it player, fighter, item, etc. that operate on the Velocity of the entity before the the transform is applied. So we no longer have the issue of clamping fighter positions all over the place!

I also removed the `move_in_arch` system in favor of using forces and velocities to create the arch. I think this is cleaner, but that's subjective, so we could add move in arch again after this PR if you wanted. Though I think we might want to wait to do any more work in that direction until we figure out how we'll handle our collisions/coordinates/y-movement: #17.

### Attack and Damage

I made the damage system more generic, adding a `Health` component, a `Damageable(bool)` component and `DamageEvent` in `damage.rs`. I changed the player stats to record a `max_health`, so that we always know how much health they can go up to, and their actual health is stored in the `Health` component. This will allow us to add health to all kinds of stuff, such as destructible scenery, that can be damaged by attacks, without having to specialize the attack system for what it is attacking.

The attack system doesn't change much, but is now generic over anything with `Health` and `Damageable` and attacks now have a `velocity` that can be used to figure out which direction and how hard an attack should push what was attacked, if applicable. This is used for knockback on the player, but would be ignored if a wall was attacked for instance.

Damage events are emitted when an attack lands, and this event is used by the knockback collector system to put the player into a knockback state when they get attacked.

zicklag:
I just discovered that the whole custom command queues thing I made isn't even necessary! Bevy automatically flushes the command queues at the end of every `SystemStage` by default, unless you explicitly make a stage that does _not_ flush the queue.

That's rather convenient!

odecay:
> I just discovered that the whole custom command queues thing I made isn't even necessary! Bevy automatically flushes the command queues at the end of every `SystemStage` by default, unless you explicitly make a stage that does _not_ flush the queue.
> 
> That's rather convenient!
> 
> Pushed an update that switches back to normal `Commands`.

yep this was the intent of splitting stuff up into stages :+1: 
glad we can avoid the custom commands.

I think I'm starting to recover a bit from being sick so I should have a chance to review this and catch up with the ongoing discussion.

odecay:
I intend to review code but before that 

> It uses the flow of: Collect Transition Intents ( from user input, enemy AI, or things like collisions ) → Evaluate Apply State Transitions → Run State Handlers.

I like this general design intent :+1: 

> Attack and Damage

I like the damage changes, will be a useful addition for when we get to breakable items.
Other than that the attack changes seem relatively minor in their design implications which is good for this PR.

> Projectiles & Lifetime

This is what I had originally implemented the lifetime for, (was called AttackLifetime before) but moved it to be projectile specific because that seemed to be the only place it made sense design wise after introducing the frame based melee attack changes.
I see it could be useful as a more general thing, but it was also used as sortof a marker component for projectiles before. Maybe if we want to use it generally it would make sense to move it out of the attacks.rs file (not sure if you did this already because I haven't looked at code yet)

>Despawn

The `DespawnMarker` was introduced because of an issue that was occurring due to parallelism in system schedule + commands application. What was happening was that one system in a stage (the systems were all in the same stage) would add a component to an entity, and another would despawn the same entity, then when the end of stage was reached, commands would apply and non-deterministically depending on the order, if despawn was applied first the add component command would cause a panic. Thus solution was to add the `DespawnMarker` component and move actual despawning of any entity that could be modified to the `Last` Stage. This is basically a contrivance to get around the panic, and if we can solve it with our scheduling/ordering that is definitely preferable, although we should keep our eye on it to make sure it isn't reintroduced by having systems which can add components to entities and systems which can despawn the same entities within the same stage.

>Removed Stages

I am fine with simplifying stages, but also with the understanding that if we find the need for more granular stages in future we will add as needed. I do agree that its annoying to figure out where a stage is added though, especially if stages are added outside of `main.rs`.

I'll return with more feedback as I go over code.

zicklag:
> The DespawnMarker was introduced because of an issue that was occurring due to parallelism in system schedule + commands application. 

Oh, yeah, I forgot to come back to this after discovering that command flushing happened between stages and not at the end of the frame. The original design makes a lot more sense after realizing that.

If you want I can add it back, or we could wait until we have another problem, because the re-organization of the stages might have fix any problems from earlier.

>  but also with the understanding that if we find the need for more granular stages in future we will add as needed.

Absolutely. I just didn't want to keep extra stages around if they weren't helping yet.

> I'll return with more feedback as I go over code.

:+1: :+1:

src/fighter_state.rs
```diff
@@ -0,0 +1,747 @@
...
+        (
+            &ActionState<PlayerAction>,
+            &mut StateTransitionIntents,
+            &Inventory,
+            &Stats,
+        ),
+        With<Player>,
+    >,
+) {
+    for (action_state, mut transition_intents, inventory, stats) in &mut players {
```
odecay, line 269: bevy 0.8 query syntax lesgoo

src/fighter_state.rs
```diff
@@ -0,0 +1,747 @@
...
+                    true,
+                ));
+            } else {
+                transition_intents.push_back(StateTransition::new(
+                    Grabbing,
+                    Grabbing::PRIORITY,
+                    true,
+                ));
+            }
+        }
```
odecay, line 294: intent triggering on input seems quite nice, liking how this looks :+1:

src/fighter_state.rs
```diff
@@ -0,0 +1,747 @@
...
+                    velocity: direction * stats.movement_speed,
+                },
+                Moving::PRIORITY,
+                false,
+            ));
+        }
+    }
+}
+
+/// Look for attacks that have contacted a figher and knock them back
```
odecay, line 312: Good in this PR but leaving a note that future design intent will require not all attacks to apply knockback.

src/fighter_state.rs
```diff
@@ -0,0 +1,747 @@
...
+        );
+    }
+}
+
+// Initiate any transitions from the flopping state
+fn transition_from_attacking(
+    mut commands: Commands,
+    mut fighters: Query<(Entity, &mut StateTransitionIntents, &Attacking)>,
+) {
+    'entity: for (entity, mut transition_intents, flopping) in &mut fighters {
```
odecay, line 370: I am unfamiliar with the `'entity:` syntax here. Is it a lifetime thing?

src/fighter_state.rs
```diff
@@ -0,0 +1,747 @@
...
+        Entity,
+        &mut Animation,
+        &mut Transform,
+        &mut LinearVelocity,
+        &Facing,
+        &Stats,
+        &Handle<FighterMeta>,
+        &mut Attacking,
+        Option<&Player>,
+        Option<&Enemy>,
```
odecay, line 457: Could potentially use an Or filter

src/fighter_state.rs
```diff
@@ -0,0 +1,747 @@
...
+    for (entity, mut animation, mut velocity) in &mut fighters {
+        // Start playing the dying animation if it isn't already
+        if animation.current_animation.as_deref() != Some(Dying::ANIMATION) {
+            **velocity = Vec2::ZERO;
+            animation.play(Dying::ANIMATION, false);
+
+        // When the animatino is finished, despawn the fighter
+        } else if animation.is_finished() {
+            commands.entity(entity).despawn_recursive();
+        }
```
odecay, line 650: In regards to our `DespawnMarker` discussion, I'm thinking that if `dying` is the highest priority state and exclusive, and no other states despawn fighters we will not have the commands applied to non-existent entity issue at least for Fighters. Could still be an issue for items or other things but unsure.

src/fighter_state.rs
```diff
@@ -0,0 +1,747 @@
...
+
+        // Tick the knock-back timer
+        knocked_back.timer.tick(time.delta());
+
+        // Set our figher velocity to the knock back velocity
+        **velocity = knocked_back.velocity;
+    }
+}
+
+/// Update dying players
```
odecay, line 636: "Update dying fighters" I think?

src/fighter_state.rs
```diff
@@ -0,0 +1,747 @@
...
+
+            // If the item asset isn't loaded yet
+            } else {
+                // This shouldn't happen because we make sure our assets are loaded before we start
+                // the game. But just in case, we'll want to know about it if it does happen.
+                warn!("Thrown item not spawned because it has not been loaded yet");
+            }
+        }
+
+        // Throwing is an "instant" state, that is removed at the end of every frame
```
odecay, line 705: Will probably want this to have an animation and not be instant in future.

src/lifetime.rs
```diff
@@ -0,0 +1,28 @@
...
+    time: Res<Time>,
+) {
+    for (entity, mut lifetime) in &mut entities {
+        lifetime.tick(time.delta());
+
+        if lifetime.finished() {
+            commands.entity(entity).despawn_recursive();
+        }
+    }
+}
```
odecay, line 28: Ah this is where the lifetime ticking and cleanup moved to :+1: seems good.

src/ui.rs
```diff
@@ -43,6 +50,22 @@ impl Plugin for UIPlugin {
...
+}
+
+// Transition game out of paused state
+fn unpause(mut commands: Commands, input: Query<&ActionState<MenuAction>>) {
+    let input = input.single();
+    if input.just_pressed(MenuAction::Pause) {
+        commands.insert_resource(NextState(GameState::InGame));
+    }
+}
+
```
odecay, line 68: lot of nice small refactors cleaning up main.rs :+1:

src/movement.rs
```diff
@@ -1,293 +1,134 @@
...
+                    .label(VelocitySystems)
+                    .after(ForceSystems)
+                    .run_in_state(GameState::InGame)
+                    .with_system(
+                        // Here we add a chain of systems that act as constraints on movements, ending
+                        // the chain with the velocity system itself which applies the velocities to the
+                        // entities.
+                        update_left_movement_boundary
+                            .chain(constrain_player_movement)
+                            .chain(velocity_system),
```
odecay, line 50: Nice approach :+1:

src/movement.rs
```diff
@@ -1,293 +1,134 @@
...
-            arc.angle.to_radians().sin(),
-        )
-        // .normalize()
-            * arc.radius;
+#[derive(Component, Deref, DerefMut, Default, Clone, Copy)]
+pub struct Force(pub Vec2);
 
-        transform.translation.x = arc.origin.x + dir.x;
-        transform.translation.y = arc.origin.y + dir.y;
+// Aplies forces to linear velocities
```
odecay, line 102: Applies

src/movement.rs
```diff
@@ -1,293 +1,134 @@
...
-        let rotation_factor = match rotate.to_right {
-            true => -1.,
-            false => 1.,
-        };
+#[derive(Component, Deref, DerefMut, Default, Clone, Copy)]
+pub struct Torque(pub f32);
 
-        transform.rotation *=
-            Quat::from_rotation_z(rotation_factor * rotate.speed * time.delta_seconds());
+// Aplies torques to angular velocities
```
odecay, line 114: Applies

src/movement.rs
```diff
@@ -305,110 +146,86 @@ pub fn update_left_movement_boundary(
...
-
-            if any_enemy_behind_stop_point {
-                for (location, movement) in player_movements.iter_mut() {
-                    if let Some(movement) = movement.as_mut() {
-                        // Can be simplified, but it's harder to understand.
-                        if location.x + movement.x > *current_stop_point {
-                            movement.x = 0.;
-                        }
-                    }
+    // Identify the current stop poing
```
odecay, line 166: stop point

src/movement.rs
```diff
@@ -1,293 +1,134 @@
...
-                Some(knockback.direction * time.delta_seconds()),
-            )
-        })
-        .collect::<Vec<_>>();
-
-    let player_dirs = player_movement_clamper.clamp(player_movements);
+/// An entity's linear velocity.
+///
+/// This is similar to the velocity you would set in a physics simulation, but in our case we use a
+/// simple constraints system instead of actual physics simulation.
```
odecay, line 61: appreciate the doc comments here

src/movement.rs
```diff
@@ -1,293 +1,134 @@
...
-
-            //Set the player state and direction
-            if dir.x < 0. {
-                facing.set(Facing::Left);
-            } else if dir.x > 0. {
-                facing.set(Facing::Right);
-            }
+/// An entity's angular velocity.
+///
+/// A positive value means a clockwise rotation and a negative value means couter-clockwise.
```
odecay, line 75: :+1:

src/movement.rs
```diff
@@ -1,293 +1,134 @@
...
-pub struct MoveInDirection(pub Vec2);
-
-// (Moving) bondary before which, the players can't go back.
-#[derive(Component)]
-pub struct LeftMovementBoundary(f32);
-
-impl Default for LeftMovementBoundary {
-    fn default() -> Self {
-        Self(-LEFT_BOUNDARY_MAX_DISTANCE)
+/// Plugin handling movement and rotation through velocities and torques.
```
odecay, line 15: I was pre-annoyed at the thought of having to work with torques, but after reading further seems like that really only comes into play in application of velocity systems so :+1:

odecay (commented):
Massive undertaking, nice job. Overall looking good, left comments.

src/enemy_ai.rs
```diff
@@ -0,0 +1,134 @@
...
+    mut commands: Commands,
+) {
+    for (entity, transform, stats, target, mut facing, mut intents) in &mut query {
+        let position = transform.translation.truncate();
+        let velocity = (target.position - position).normalize() * stats.movement_speed;
+
+        // If we're close to our target
+        if position.distance(target.position) <= target.attack_distance {
+            // Note that the target includes an offset, so this can still not point to the
+            // player.
```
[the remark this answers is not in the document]
zicklag: That's still valid right? I just copied the comment, I think from #198.

src/fighter_state.rs
```diff
@@ -0,0 +1,747 @@
...
+}
+
+/// Component indicating the player is dying
+#[derive(Component, Reflect, Default, Debug)]
+#[component(storage = "SparseSet")]
+pub struct Dying;
+impl Dying {
+    pub const PRIORITY: i32 = 1000;
+    pub const ANIMATION: &'static str = "dying";
+}
```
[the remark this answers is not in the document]
zicklag: Since this moves towards a more dynamic fighter state list than before, the idea is that we don't necessarily know all of the possible animations that might exist at compile time, so each state is responsible for asking the fighter to play a named animation, instead of a fixed enum variant.

We'll probably want to be somewhat standardized on which animations should exist as a baseline for fighters, but this will allow mods to create new fighters with completely new states, attacks, and animations.

src/fighter_state.rs
zicklag: So far it's working very nicely and was refreshingly easy to re-build the fighter state machine with the new model, piece by piece.

It felt easy to just come up with the next thing I wanted the fighter to be able to do, and it didn't require modifying a big centralized system of state management. I just had to add a new state, decide when to trigger it, and how to handle it.

---

I'm still a little unsure about what the best practice will be as far as where to make decisions about which state to transition to. For instance, right here we have the `collect_player_actions` system making a decision, based on the fighter's inventory, what state to transition to.

This seems good in this case, but it get's a little more confusing when consider things like mid-air attacks where the same attack button might trigger a different state. Should `collect_player_actions` be responsible for making that decision? Or does it just depend on a case-by-case basis?

We'll just have to think about options as we go I think and feel out how well it's working and what might be the best practice.

src/fighter_state.rs
zicklag: :+1:

Not for this PR, but here's a thought:

I already have a `damage_velocity` field on the `Damage` component.

Maybe instead we should change `damage_velocity` to `damage_impuse` and make it a new struct like this:

```rust
struct Impulse {
    velocity: Vec2,
    time: f32,
}
```

Then if you made an attack where the damage_impulse was `Impulse::default()` which has a `time` and `velocity` of 0.0, this system would check that, and in that case it wouldn't put the player into a knockback state.

src/fighter_state.rs
zicklag: Nope, that's actually a [loop label](https://doc.rust-lang.org/stable/reference/expressions/loop-expr.html?highlight=label#loop-labels).

Look down below where we say `continue 'entity`. It allows us to continue the entity loop, and would help distinguish which loop you were `continue`-ing, if you needed to `continue` from inside a nested loop.

I needed it at one point, but apparently I took out whatever nested loop I had and it's flat now, so that isn't necessary anymore. I can take it out.

It's actually a really nice feature I've never seen in another language, but I guess we don't need it here anymore. :)

src/fighter_state.rs
zicklag: I don't think I could tell which one is present if we used an or filter, though. I need to call `player.is_some()` down below.

src/fighter_state.rs
zicklag: Ah yes. :+1:

src/fighter_state.rs
zicklag: OK, then I think, since it's so easy to implement later, if necessary, we can just wait to see if we run into issues.

src/fighter_state.rs
zicklag: Good point.

Cool thing is, that's super easy to change!

We just have to make the throwing state transition non-additive so that it takes over for the other states, and then we can be free to play an animation during the `throwing` system.

src/input.rs
```diff
@@ -5,7 +5,7 @@ use serde::Deserialize;
 pub enum PlayerAction {
     Move,
     // Attacks
-    FlopAttack,
+    Attack,
```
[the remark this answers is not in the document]
zicklag: Sounds good. I'll switch it back.

zicklag (commented):
Great review!

I left responses where applicable and I'll push an update once I get the chance.

src/enemy_ai.rs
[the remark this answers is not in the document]
odecay: yeah all good

src/fighter_state.rs
odecay: yeahp im wondering that also but it doesnt need to be addressed here.
I was thinking about it in relation to combo/chain attacks and input buffering where you want to chain attacks/play animations based on sequence of buttons within certain frame windows.

src/fighter_state.rs
odecay: Oh very interesting. Thanks for explaining, I'm happy to know about that.
If we don't need it here definitely remove it.

src/attack.rs
```diff
@@ -471,9 +74,10 @@ fn activate_hitbox(
     }
 }
 
+/// Deactivate collisions for entities with [`AttackFrames`]
 fn deactivate_hitbox(
     query: Query<(Entity, &AttackFrames, &Parent), (With<Attack>, With<Collider>)>,
-    fighter_query: Query<&Animation, With<State>>,
+    fighter_query: Query<&Animation>,
```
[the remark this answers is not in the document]
odecay: yeap that sounds good

src/fighter_state.rs
odecay: Ohh you are right I didn't actually notice that `is_player, is_enemy` variables were used again after the continue check.

src/fighter_state.rs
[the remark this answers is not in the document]
odecay: I think I'm sort pointing toward my perception that the dynamic states stuff could have been separated from the rest of the state refactor since enabling that was "part of scripting improvements" in my mind, but I think ultimately I agree with the design goal and it implicates the state refactor just as much as it does scripting so I think I dont really have a reason to object.

src/commands.rs
```diff
@@ -0,0 +1,38 @@
...
+        let entity = self.id();
+        self.commands().add(InsertDynamicCommand(
+            entity,
+            reflect_component,
+            component_data,
+        ));
+
+        self
+    }
+}
```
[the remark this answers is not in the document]
odecay: I think I may have missed how/where the requirement for dynamic components was coming in, or misattributed to scripting when its required for the states in general?

src/fighter_state.rs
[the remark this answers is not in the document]
zicklag: Yeah, in this case I think what was essential for scripting, still turns out useful to us for easily extending our own design as we move on, as long as our end design goal isn't to only ever have a fixed set of animations.

Here is one middle-ground alternative, worth considering, though:

We have a core set of animations we expect to be implemented for every fighter, we could enumerate those, but we allow custom, arbitrarily named animations also, that could be triggered by custom attacks. That way you could try to make sure every fighter implements the "Core" animations, but it wouldn't stop mods from adding new animations.

That could be good to help allow flexibility, while still enforcing some sort of standard. And having a standard is good for other things, too, for instance, if we have special effects later. If we required that all fighters were required to have the "stunned" core animation, then any time somebody invents a new mod with a new status effect like "stuck in a bubble", they would know that it would be safe to play the "stunned" animation, because all fighters have the "stunned" animation.

But if somebody made a new fighter, it would still be possible for them to have the "ultimate_smash" animation, which is non-standard and only available to that fighter.

src/commands.rs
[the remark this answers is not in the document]
zicklag: Yes the dynamic components are actually used in states, regardless and totally unrelated to scripting.

Though it does make scripted attacks _way_ easier to implement later, it is still absolutely useful and utilized in the current state system completely irrespective of any scripting effort.

It's actually kind of a good sign that you missed it, at least as far as readability state transition logic is concerned. I'm glad the new concept didn't make the state transition systems' implementations hard to understand.

I'll walk through it and why it helps.

---

When we want to express the intent to transition to a new state we do something like this:

```rust
if action_state.just_pressed(PlayerAction::Attack) {
    transition_intents.push_back(StateTransition::new(
        Attacking::default(),
        Attacking::PRIORITY,
        false,
    ));
}
```

We express the desired to add an `Attacking` component to the fighter, and we push it into the `VecDeque`.

Later in the same function we do this:

```rust
transition_intents.push_back(StateTransition::new(
    Throwing,
    Throwing::PRIORITY,
    true,
));
```

Here we express the desire to add a _different_ component to the fighter, but we are pushing the intent to the _same_ `VecDeque`!

This means we have to "erase" the component type so that we can store multiple different component types in the same `VecDeque`. This is conceptually similar to needing to do a `Vec<Box<dyn Any>>` if you want to store multiple different types in the same vector.

This type erasure is manged by the `StateTransition::new` function. It boxes the value and erases it's type while also querying the Bevy type registry to get the info needed to add the component to an entity later. And all that gets stored in the `StateTransition`, along with the `priority` and `is_additive` options.

Finally, when we are looping through state transition intents and deciding which one to transition to we do this:

```rust
if intent.priority > current_state_priority {
    // Apply the state
    intent.apply::<CurrentState>(entity, commands);
}
```

We are saying that we want to take the component stored in the `StateTransition` `intent` and add it to the fighter, even though we don't know what component that is, because it's been type erased.

After my latest commit, to get rid of the confusing commands stuff, `StateTransition::apply` looks like this:

```rust
pub fn apply<CurrentState: Component>(self, entity: Entity, commands: &mut Commands) -> bool {
    if !self.is_additive {
        commands.entity(entity).remove::<CurrentState>();
    }

    commands.add(move |world: &mut World| {
        // Insert the component stored in this state transition onto the entity
        self.reflect_component
            .insert(world, entity, self.data.as_reflect());
    });

    self.is_additive
}
```

The `reflect_component` has the type information needed to insert a new instance of the type-erased component onto the entity with the `insert` function. And we must pass it the `world`, the `entity` to add the component to, and our `Box<dyn Reflect>` which contains our actual component data.

---

That completes the workflow and allows us to store a bunch of transitions to any kind of new state in a single queue. And using components instead of an enum allows us to use the awesome Bevy component filtering syntax in our queries.

All that has "nothing" to do with scripting. But it makes it does make it trivial for a script to decide to emit a state transition intent for a new, unknown state, that is handled by the script. As long as the priority is higher than the existing state, the state will transition in the Rust system, and at that point the script state handler takes over and can control the player however it wants at that point. Voilà! You have a new attack/status effect/whatever without modifying core.

And this pattern, again, is helpful even if we only ever wrote Rust.

zicklag:
OK, pushed a few updates resolving all but the last two unresolved conversations.

src/commands.rs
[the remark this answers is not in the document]
odecay: Yep seems good. Very valuable clear explanation. I wonder if it would be good to retain this doc somewhere, seems useful in case someone wants to understand whats going on under the hood.

I caught the way we are using VecDeque but was missing the part about what was stored inside being type erased. Although I have a basic understanding of Boxing/type erasure I've never actually used it, and totally missed that we were storing an actual component in the state transition.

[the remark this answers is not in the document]
odecay: I think that about clears up the last of my questions. Probably good to merge whenever you are ready.

src/fighter_state.rs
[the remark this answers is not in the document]
odecay: Yeah I guess it depends on what we want design wise right? If we want people to make entirely custom "games" as mods then it might make sense not to have any limitations on what a base set of animations are.

It also might be nice for prototyping fighter definitions/development workflow purposes to be able to load them with only a subset of animations but I think thats a different issue.

I guess we should note this as something to come back to. Definitely don't need to change that for current PR.

odecay (approved):

src/commands.rs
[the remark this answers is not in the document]
zicklag: > I wonder if it would be good to retain this doc somewhere, seems useful in case someone wants to understand whats going on under the hood.

Great idea. I'll move it to the Wiki as a sub-page under Architecture probably, and I'll also write up an explanation of the high-level Architecture after merging.

edgarssilva:
@zicklag I really like the description on the wiki for the state machine. Are you ever considering creating a crate with it?

zicklag:
> Are you ever considering creating a crate with it?

The thought hadn't actually crossed my mind. :thinking: Interestingly, the non-game-specific code only adds up to 62 lines of code!

Still, it does seem like it could be useful as a crate, so that might be worth doing.

Issue #127, opened 2022-07-24
Items visualization


64kramsystem:
### Description

After merging #126, basic support for items will be in place, which implements the functional part, but not the visual one.

The visual part must therefore be added. This is a matter of design first, since there isn't a single way to implement it.

At a minimum (which I think it'd be a good starting point, as in "better than nothing" :joy:), we could display a bottle in the HUD, when the player has a bottle.

64kramsystem:
I'm marking this as small. It's not well-defined in general terms, but the MVP version is.

zicklag:
Yeah, it seems like showing the item in the HUD right here would be a pretty nice first pass. Eventually we probably want to show the fish holding the items:

![image](https://user-images.githubusercontent.com/25393315/180656368-7b9e5331-959c-43ba-b9f7-74670fdf696b.png)

edgarssilva:
I'd like to take on this task, I have some ideas and some changes I would like to make but first to have some input on it.
What I'm thinking of is creating an Inventory component that holds items (probably non-stackable) using the Item (this could be the old ItemBundle) struct, therefore I think it's a good idea for Item to not be a component but a simple struct and create a new one called DroppedComponent that holds an Item. 
This way the items can only be accessed via the wrapper component and logic can be treated separately instead of checking whether the entity holding the item has a transform or not. 

If you think is not a good approach or is just unnecessary work I can just store entities in the inventory, and keep the transform logic.

zicklag:
I think that using a component to store item inventory is probably a good idea. I see you opened #206 which I'll try to review as soon as I get the chance. :+1:

PR #206, opened 2022-08-05
Inventory system

edgarssilva:
Just an initial version for #127, still need to define how items are defined, if by entities or structs with "wrapper" components.

The items are correctly picked up and removed on attack/throw, though this currently leads to a bug since the key to pick up and throw an item is the same. The item gets picked up and thrown before you can lift the key, this kinda makes it hard to see the HUD.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /target
 /web-target
 .cargo
+/.idea
--- a/src/fighter.rs
+++ b/src/fighter.rs
@@ -3,6 +3,7 @@ use bevy_rapier2d::prelude::CollisionGroups;
 use rand::prelude::SliceRandom;
 use serde::Deserialize;
 
+use crate::item::Item;
 use crate::{
     animation::{AnimatedSpriteSheetBundle, Animation},
     camera::YSort,
@@ -14,7 +15,6 @@ use crate::{
     movement::LinearVelocity,
     player::Player,
 };
-use crate::item::Item;
 
 /// Bundle added to a fighter stub, in order to activate it.
 #[derive(Bundle)]
@@ -114,4 +114,3 @@ impl ActiveFighterBundle {
         commands.entity(entity).insert_bundle(active_fighter_bundle);
     }
 }
-
--- a/src/fighter.rs
+++ b/src/fighter.rs
@@ -3,7 +3,7 @@ use bevy_rapier2d::prelude::CollisionGroups;
 use rand::prelude::SliceRandom;
 use serde::Deserialize;
 
-use crate::item::Item;
+use crate::metadata::ItemMeta;
 use crate::{
     animation::{AnimatedSpriteSheetBundle, Animation},
     camera::YSort,
@@ -48,7 +48,7 @@ pub struct Stats {
 ///
 /// A player may be holding one item
 #[derive(Component, Clone, Debug, Default, Deref, DerefMut)]
-pub struct Inventory(pub Option<Item>);
+pub struct Inventory(pub Option<ItemMeta>);
 
 impl Default for Stats {
     fn default() -> Self {
--- a/src/fighter.rs
+++ b/src/fighter.rs
@@ -10,10 +10,11 @@ use crate::{
     damage::{Damageable, Health},
     enemy::Enemy,
     fighter_state::{Idling, StateTransitionIntents},
-    metadata::{FighterMeta, ItemMeta},
+    metadata::FighterMeta,
     movement::LinearVelocity,
     player::Player,
 };
+use crate::item::Item;
 
 /// Bundle added to a fighter stub, in order to activate it.
 #[derive(Bundle)]
@@ -47,7 +48,7 @@ pub struct Stats {
 ///
 /// A player may be holding one item
 #[derive(Component, Clone, Debug, Default, Deref, DerefMut)]
-pub struct Inventory(Option<Handle<ItemMeta>>);
+pub struct Inventory(pub Option<Item>);
 
 impl Default for Stats {
     fn default() -> Self {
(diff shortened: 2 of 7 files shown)


zicklag (commented):
I like this. Moving item storage to a component is something I felt we should do and the hud display works well.

I do wonder whether or not we should have multiple item slots in inventory, though. I think the player can probably only carry one item at a time, at least of the kind that can be thrown. If we have other different kinds of items later, like a shield hat or something like that, they should probably be stored in a different component, so I think having just an `Option` for the inventory storage should be good. But I'd have to get feedback from @odecay or @erlend-sh on whether or not that's right.

The other thing is that I did touch this code ( like I touched almost all the gameplay code ) in #203, which if we merge adds an inventory component similar to this. I still need to get feedback on that PR, but even if that did get merged we can still use the HUD code from here and it would only change slightly.

So I can't say anything definitive until I get some feedback on what the plan is with #203, but good job!

edgarssilva:
Thanks, wasn't aware of that PR, probably should have worked on top of it. After #203 merge, I'll try and refractor it and fix conflicts.

zicklag:
#203 was kind of a big experiment and I'm not sure what the feedback on it will be yet, so you _might_ want to wait to refactor on top of it, but it's up to you.

Also if you have any thoughts at all on how things are implement #203 let me know!

edgarssilva:
> #203 was kind of a big experiment and I'm not sure what the feedback on it will be yet, so you _might_ want to wait to refactor on top of it, but it's up to you.
> 
> Also if you have any thoughts at all on how things are implement #203 let me know!

I was intending to wait for the merge, I took a quick look mostly at the stuff related to items.

In the case of how many items the player can have, I would think something like 3 to 5 is a nice spot, for the case of bottles you could make it so the player can grab up to 3 before throwing one. This is more of a gameplay design choice so not sure.

So far look great I'll leave some thoughts in the PR.

64kramsystem (approved):
Nice one, fits well. Looks good to me! :+1:

edgarssilva:
Since #203 has been merged, I'm taking a look at this. 
I've been thinking of changing Item and ItemBundle into just Item that has a name and a handle to metadata. Is that ok?

zicklag:
> I've been thinking of changing Item and ItemBundle into just Item that has a name and a handle to metadata. Is that ok?

@odecay do you have a preference?

My gut reaction is to keep `Handle<ItemMeta>` as a separate component, because for everything we've done so far we can do a query like this to get the metadata: `Query<&Handle<[SomeKindOf]Meta>>`.

I feel like it's just about right how it is, but if you guys think it should change I'm not dead set on it.

odecay:
> Item that has a name and a handle to metadata. Is that ok?

So removing all the other properties?

Can you give a code snippet/psuedocode for what it would look like either way?

edgarssilva:
Just pushed some changes, to answer @odecray basically just take the components in the bundle and put them in the Item component,  you can check the diff. 
I wasn't thinking much has a `Hande<ItemMeta>` being a component but what @zicklag pointed out makes sense. 
If you could take a review and see if you want me to roll back that change.

odecay:
Hmm okay taking a look, I feel like I have a slight preference against moving those to fields instead of their own components because it means we cant query against them.

zicklag:
We could maybe clone the `ItemMeta` from the handle into the the `Item` struct, so that we would't have to fetch the handle data every time we wanted it.

Since we are loading all the level assets before the level is spawned, now, that shouldn't be a problem, and it would save us having to do extra asset lookups when we know the data is already there.

We might want to do that for fighters, too actually. It's already led to questions on what to do when the handle isn't loaded, when in reality it should always be loaded once the game starts anyway.

---

Also, we should keep the `Name` component separate, because that's actually used by the world inspector to label the entity.

odecay:
> We could maybe clone the `ItemMeta` from the handle into the the `Item` struct, so that we would't have to fetch the handle data every time we wanted it.
> 
> Since we are loading all the level assets before the level is spawned, now, that shouldn't be a problem, and it would save us having to do extra asset lookups when we know the data is already there.
> 
> We might want to do that for fighters, too actually. It's already led to questions on what to do when the handle isn't loaded, when in reality it should always be loaded once the game starts anyway.
> 
> Also, we should keep the `Name` component separate, because that's actually used by the world inspector to label the entity.

I was thinking about a pattern like loading from asset once on spawn/hotreload and adding components to the entity that correspond to data from YAML. Then we could query for individual fields/properties and not have to think about assets when writing gameplay logic.

edgarssilva:
> Hmm okay taking a look, I feel like I have a slight preference against moving those to fields instead of their own components because it means we cant query against them.

I'll make the necessary changes tomorrow. 

And yeah I think it would be much nicer  to have direct access to the meta instead of handles.

zicklag:
> And yeah I think it would be much nicer to have direct access to the meta instead of handles.

Yeah, I think that makes sense. That's actually how the `Stats` component works on fighters. We spawn the fighter with just a handle, and then we populate the fighter components afterward. See [`load_fighter()`](https://github1s.com/fishfolks/punchy/blob/HEAD/src/loading.rs#L427-L428) for reference.

The caveat is it takes an extra system to handle hot reloads because we aren't checking for new data every frame by loading handles. Still, its worth the trade-off I think. See how the fighters are hot reloaded in [`hot_reload_fighters()`](https://github1s.com/fishfolks/punchy/blob/HEAD/src/loading.rs#L457-L458).

If we have smaller components that makes sense to create from the metadata, we can do that, otherwise we probably just want to clone the entire metadata struct and stick it directly on the fighter/item so that we have all of it available as a component.

edgarssilva:
Just reverted back to the ItemBundle, just made a small change in the Inventory instead of storing a Handle it stores the actual ItemMeta that is loaded on the grab item system.

zicklag (approved):
Didn't test it, but code looks fine to me. :+1:

64kramsystem (approved):

PR #219, opened 2022-08-10
making clear to the player

DRuppFv:
Just a very simple PR to say that I have free time to dedicate myself to the project again...

I've tried calling it "throw/grab", but I don't know how to make the .ftl file ignore a "/", so if you know, feel free to fix it.


edgarssilva (approved):
That seems about right, I'll merge it.

PR #221, opened 2022-08-10
Load attack properties from assets

odecay:
This loads basic attack properties from YAML fighter definition files.
AttackFrames, hitbox size and offset relative to fighter are currently loaded.
I've also added a stationary `Punching` attack, used by enemies AI and player. Flop attack still exists but is currently disabled with the intent to re enable after input buffer + attack chain system work.

Fighters now have attack frames which better fit the differing lengths of their attack animations, though the placement and sizing of hitboxes are currently not tuned to match the sprites, though easily modifiable via fighter definitions if need arises.

Direct follow up work to this is additional attack properties loaded from YAML, but this should be left till after other basic attack work is in place.

- Knockback(whether/how much it moves the character hit)
- Hitstun(how long the enemy takes to recover and be able to perform actions when hit),
- Movement (how much it moves the character performing the attack),
- Invincibility/hyperarmor (whether the attack should be able to be interrupted).

### DIFF
--- a/assets/fighters/bandit/bandit.fighter.yaml
+++ b/assets/fighters/bandit/bandit.fighter.yaml
@@ -36,6 +36,15 @@ spritesheet:
     attacking:
       frames: [14, 17]
 
+attack:
+  name: "punch"
+  frames:
+    startup: 1
+    active: 2
+    recovery: 3
+  hitbox: [12, 12]
+  hitbox_offset: [24, 0]
+
 audio:
   effects:
     attacking:
--- a/src/attack.rs
+++ b/src/attack.rs
@@ -2,10 +2,12 @@ use bevy::{hierarchy::DespawnRecursiveExt, math::Vec2, prelude::*, reflect::Refl
 use bevy_rapier2d::prelude::*;
 use iyes_loopless::prelude::*;
 
+use serde::Deserialize;
+
 use crate::{
     animation::Animation,
-    consts::{ATTACK_HEIGHT, ATTACK_WIDTH},
     damage::{DamageEvent, Damageable, Health},
+    metadata::FighterMeta,
     GameState,
 };
 
@@ -42,28 +44,31 @@ pub struct Attack {
 ///
 /// Must be added to an entity that is a child of an entity with an [`Animation`] and an [`Attack`]
 /// and will be used to spawn a collider for that attack during the `active` frames.
-#[derive(Component)]
+/// Each field is an index refering to an animation frame
+#[derive(Component, Debug, Clone, Copy, Deserialize)]
 pub struct AttackFrames {
     pub startup: usize,
     pub active: usize,
     pub recovery: usize,
 }
 
-/// Activate collisions for entities with [`AttackFrames`]
 fn activate_hitbox(
     attack_query: Query<(Entity, &AttackFrames, &Parent), Without<Collider>>,
-    animated_query: Query<&Animation>,
+    fighter_query: Query<(&Animation, &Handle<FighterMeta>)>,
     mut commands: Commands,
+    fighter_assets: Res<Assets<FighterMeta>>,
 ) {
     for (entity, attack_frames, parent) in attack_query.iter() {
-        if let Ok(animation) = animated_query.get(**parent) {
+        if let Ok((animation, fighter_meta)) = fighter_query.get(**parent) {
             if animation.current_frame >= attack_frames.startup
                 && animation.current_frame <= attack_frames.active
             {
-                //TODO: insert Collider based on size and transform offset in attack asset
-                commands
-                    .entity(entity)
-                    .insert(Collider::cuboid(ATTACK_WIDTH * 0.8, ATTACK_HEIGHT * 0.8));
+                if let Some(fighter_data) = fighter_assets.get(fighter_meta) {
+                    commands.entity(entity).insert(Collider::cuboid(
+                        fighter_data.attack.hitbox.x,
+                        fighter_data.attack.hitbox.y,
+                    ));
+                }
             }
         }
     }
--- a/src/fighter_state.rs
+++ b/src/fighter_state.rs
@@ -55,7 +55,9 @@ impl Plugin for FighterStatePlugin {
                     .after(FighterStateCollectSystems)
                     .run_in_state(GameState::InGame)
                     .with_system(transition_from_idle)
-                    .with_system(transition_from_attacking)
+                    .with_system(transition_from_flopping)
+                    .with_system(transition_from_punching)
+                    .with_system(transition_from_ground_slam)
                     .with_system(transition_from_knocked_back)
                     .into(),
             )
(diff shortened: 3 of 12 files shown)


assets/fighters/bandit/bandit.fighter.yaml
```diff
@@ -36,6 +36,17 @@ spritesheet:
     attacking:
       frames: [14, 17]
 
+# maybe attacks should be defined in their own files and then referenced in fighters?
+# but there is some inherent connection to the attack frames in each fighters unique spritesheet also
```
zicklag, line 40: To me it feels right in the same file.

I've modded the old Star Wars Battlefront game which used _way_ to many files with dependencies on each-other and it can be a real pain when you end up with a bunch of files that are _too_ coupled and end up feeling that they should just be one big file.

Not to say we're anywhere near that yet, but I think it's probably sensible to allow the fighter attacks to be tied closely to the fighters in the same file.

src/attack.rs
```diff
@@ -47,28 +49,440 @@ pub struct Attack {
...
+//         }
+//     }
+// }
+
 fn activate_hitbox(
     attack_query: Query<(Entity, &AttackFrames, &Parent), Without<Collider>>,
-    animated_query: Query<&Animation>,
+    fighter_query: Query<(&Animation, &Handle<FighterMeta>)>,
     mut commands: Commands,
+    //maybe change to using attack meta? or atatcks defined in their own files but still loaded via
```
zicklag, line 470: Maybe...but I'd probably wait until we have something other than fighters that actually wants to use this attack system.

src/fighter_state.rs
```diff
@@ -535,6 +579,7 @@ fn attacking(
 
         // Do a forward jump thing
         //TODO: Fix hacky way to get a forward jump
+        //should moving and any attack that requires movement be a non exclusive state?
```
zicklag, line 582: Not 100% sure I understand what this means, but my gut reaction is that movement made during an attack should be manged by an exclusive attack state most of the time. Unless you can literally walk while attacking without the attack interfering with the walking movement at all.

src/fighter_state.rs
```diff
@@ -554,10 +599,111 @@ fn attacking(
...
+                // Play attack sound effect
+                if let Some(effects) = fighter.audio.effect_handles.get(Punching::ANIMATION) {
+                    let fx_playback = AnimationAudioPlayback::new(
+                        Punching::ANIMATION.to_owned(),
+                        effects.clone(),
+                    );
+                    commands.entity(entity).insert(fx_playback);
+                }
+            } else {
+                //what should we actually do if we cant read attack metadata from fighter?
```
zicklag, line 698: Technically this shouldn't happen because we have logic to make sure all assets are loaded before starting gameplay. We could maybe panic with a message saying that much so that we know if we accidentally break pre-loading assets maybe?

zicklag (commented):
Did a quick review of the code and looks good. Left a few comments on your comments.

src/fighter_state.rs
odecay: Comment should have read "additive state" but I honestly don't really remember what I was getting at with this one either.

zicklag:
> Design wise I'm starting to feel like we might want to have different attacks have different properties, but be applied in a limited number of attack states.

If we end up with scripts defining attacks, then it might make sense to have free-form attack metadata, and it would be the responsibility of the attack state handler system to parse the attack options applicable to that particular attack state.

That way each attack could be configured in whatever way made sense for that attack.

That's just a quick, thought, I haven't thought carefully about it yet.

src/fighter_state.rs
zicklag: `non-exclusive` sounds nice too. :thinking: I could almost go with either terminology. :)

src/fighter_state.rs
odecay: So this is the correct way to query the metadata then? Shouldn't have to worry about the handle not returning something?

zicklag: Yeah. The handle should always return or we've done something wrong in the loader.

So we could either `.expect("Handle not loaded")` or something like that, or we could `warn!("Skipping attack due to un-loaded fighter meta")`.

I'm on the fence between the two. If we panic, we won't accidentally mis the log message, but it might crash the game for a user who doesn't know what to do about it, and it might only cause a 1-frame problem anyway and it could have been safely ignored.

If we don't panic, we face the possibility of not seeing the warning during development, and being more likely to ship a release of the game with a bug in it.

Difficult trade-off to decide between. :man_shrugging:

odecay:
> > Design wise I'm starting to feel like we might want to have different attacks have different properties, but be applied in a limited number of attack states.
> 
> If we end up with scripts defining attacks, then it might make sense to have free-form attack metadata, and it would be the responsibility of the attack state handler system to parse the attack options applicable to that particular attack state.
> 
> That way each attack could be configured in whatever way made sense for that attack.
> 
> That's just a quick, thought, I haven't thought carefully about it yet.

The properties I'm thinking of would be things like:

- Knockback(whether/how much it moves the character hit)
- Hitstun(how long the enemy takes to recover and be able to perform actions when hit), 
- Movement (how much it moves the character performing the attack), 
- Invincibility/hyperarmor (whether the attack should be able to be interrupted).

assets/fighters/bandit/bandit.fighter.yaml
odecay: Yeah, I think continuing like this probably makes sense.

zicklag:
Oh, I see that makes sense. Those are attributes as far as what the attack does to the receiver of the damage. I was thinking of attributes effecting how the attack is performed.

Yeah, those attributes make a lot of sense to me.

odecay:
> Oh, I see that makes sense. Those are attributes as far as what the attack does to the receiver of the damage. I was thinking of attributes effecting how the attack is performed.
> 
> Yeah, those attributes make a lot of sense to me.

Any examples/ideas of the sorts of things you are thinking of?

zicklag:
I just meant like if we have different kind of attacks like a ground-pound. It could be defined in Rust or in a script, either way. Then when you add that attack to a fighter you'd want to be able to configure things about the ground pound, like how big the damage radius is, how much damage it does, etc.

But there could be any number of different kind of attacks, such as punch, flop, etc. And those wouldn't have `damage_radius` they would just have `damage`, so we might need to allow a flexible attack options field at some point.

That might be somewhat relevant really soon actually, because already we've got the stomp that the boss is going to be doing, and then we'll have the slingshot attack of the other fighters, but we don't have a great way yet to say _which_ of these attacks the fighter will do when they press a certain button.

odecay:
For sure those all make sense.
I was also thinking about both those cases as well. Soon each fighter will have multiple attack types as well.
Its hard to say exactly where the line is for which attacks should be represented as their own state + transition systems and which should be loaded as properties. (I think thats what we are talking about here) 
We will probably figure out what designs map best as we progress.

odecay:
This is all I intend to put in this PR for now.
Will probably have effect on #222

edgarssilva:
Merged in #222.

edgarssilva:
Sorry about that closed the wrong PR.

zicklag (approved):
Looks good. We'll just have to resolve conflicts. If you want, I can rebase this onto the changes from #222 that were just merged.

I think we can take out the `boss` flag, now that we have an attack name field which we can use to decide which attack the fighter triggers.

odecay:
oof rebase + rebase. 
The boss AI is currently not functional but I really don't want to keep rebasing this so gonna put it in and address that in follow up.

odecay:
bors merge

PR #222, opened 2022-08-11
Rebase Boss MVP by @edgarssilva

zicklag:
Rebases #200 onto the latest master, with a couple subjective tweaks:

- Make boss flag in YAML a `bool` instead of an `Option<bool>`
- Impl HasLoadProgress for `bool` instead of ignoring the `boss` field
  in the metata struct.

Otherwise this is pretty much all just @edgarssilva's code.

@edgarssilva , you can push to this branch to do anything else you want to do with this before merging, if anything.


edgarssilva (approved):
Seems great thanks for the help.
Also, the attack logic isn't great and can lead to odd behavior but I didn't worry too much since we are changing it later.

odecay:
I think we are supposed to be using bors to merge stuff now, and remember to make the formatting checks pass also

edgarssilva:
> I think we are supposed to be using bors to merge stuff now, and remember to make the formatting checks pass also

Ho your right sorry, totally went over my head.

zicklag:
You can require bors to pass before merging on the branch protection rules, which would prevent this from accidentally happening in the future. I've almost clicked it a couple of times before remembering I was supposed to tell bors to merge it:

image

edgarssilva:
> You can require bors to pass before merging on the branch protection rules, which would prevent this from accidentally happening in the future. I've almost clicked it a couple of times before remembering I was supposed to tell bors to merge it:

Done, also should I enable "Require a pull request before merging" so that also doesn't happen accidentally.

PR #223, opened 2022-08-11
Allow Configuring Debug Tools at Runtime

### LINKED ISSUE #144: Remove `debug` Feature In Favor of Runtime Configuration

zicklag:
Closes #144.

The `Reflect` derives are used as a more universal way to get support for inspecting them with the Bevy Egui world inspector. 

We will need to derive reflect and register the types for scripting so we might as well not do that and derive `Inspectable`.

This also clears out some of the remaining system scheduling that was done in main.rs and should have been in a plugin.

Now you can show the debug tools menu with F12 and configure the visibility of collision boxes and the world inspector.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,17 +48,16 @@ directories = "4.0.1"
 async-channel = "1.6.1"
 once_cell = "1.13.0"
 
-# Waiting for updates to Bevy 0.8
+# Debug tools
+bevy-inspector-egui = { version = "0.12.1" }
+bevy-inspector-egui-rapier = { version = "0.5.0", features = ["rapier2d"] }
 bevy_mod_debugdump = { version = "0.5.0", optional = true }
-bevy-inspector-egui = { version = "0.12.1", optional = true }
-bevy-inspector-egui-rapier = { version = "0.5.0", optional = true, features = ["rapier2d"] }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 web-sys = { version = "0.3", features = ["Window","Location","Storage"] }
 
 [features]
 default = []
-debug = ["bevy_mod_debugdump", "bevy-inspector-egui", "bevy-inspector-egui-rapier"]
 schedule_graph = ["bevy_mod_debugdump"]
 
 # Enable optimizations for dependencies but not for our code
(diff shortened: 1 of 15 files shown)


64kramsystem (approved):

edgarssilva (approved):
Looks good and seems to work fine. 
Just a quick note I'm getting a flashing collision box not sure if that's from the inspector or the actual game.

zicklag:
Yeah, I've seen that before too. I think it's been there for a while, but I'm not sure what causes it. We should eventually figure that out.

odecay:
I feel like I'm missing something, I cant get this to trigger when I hit f12.

64kramsystem:
> I feel like I'm missing something, I cant get this to trigger when I hit f12.

Nothing displays also when I tap F12.

edgarssilva:
You have to provide the options, for me, I'm using `cargo run -- -d` not sure if that's the simpler one.

zicklag:
Yeah, you have to enable the debug tools on the commandline or ( on the web ) in the query string. The idea is to prevent adding plugin systems that won't be used when a normal user is playing the game.

Granted we might not have to worry about those systems impacting performance, so we might want to just enable them all the time?

edgarssilva:
Maybe we could check if it's a release build or not? Works either way for me.

zicklag:
Well, we still want to allow debug tools on release builds because modders will have the release build and will want to be able to access the debug tools.

odecay:
> You have to provide the options, for me, I'm using `cargo run -- -d` not sure if that's the simpler one.

Ah perfect, I missed that.

PR #224, opened 2022-08-12
Fix rapier debug render

### LINKED ISSUE #132: Returning to main menu in debug mode turns off rapier debug draw plugin
### Description

When returning to the main menu either by pausing or allowing both player characters to die, on restarting the game/level through the menu, the rapier debug drawing features are no longer loaded.

This makes it necessary to close and re run the program when doing certain types of debugging/development as opposed to continuing session through use of menu.

### Expected Behavior

Expect rapier debug features to remain loaded after restarting through menu.

zicklag:
Fixes #132.

Found out that #132 was caused by the fact that the debug lines plugin that rapier uses to manage debug rendering requires that 4 entities be kept alive to contain the debug lines.

The issue is when we try to reset the world by deleting all entities other than the camera, those entities get deleted, too. And we can't filter those entities out because their distinguishing component is private. Also, there isn't a way to re-create those entities after adding the plugin.

This works around the issue by implementing a custom debug lines renderer for Rapier that uses Egui to draw the lines instead.

It's simple and mostly copied from rapier's plugin.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -112,7 +112,7 @@ fn main() {
         .add_plugin(PlatformPlugin)
         .add_plugin(LocalizationPlugin)
         .add_plugin(LoadingPlugin)
-        .add_plugin(RapierPhysicsPlugin::<NoUserData>::pixels_per_meter(100.0))
+        .add_plugin(RapierPhysicsPlugin::<NoUserData>::default())
         .add_plugin(InputManagerPlugin::<PlayerAction>::default())
         .add_plugin(InputManagerPlugin::<MenuAction>::default())
         .add_plugin(AttackPlugin)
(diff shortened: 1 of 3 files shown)


64kramsystem (approved):

zicklag:
> Also figured out that the flashing boxes are the enemies' hitboxes, we might wanna check that.

Yeah we should probably create an issue for it.

> If I'm not mistaken the flop attack staying in place is due to the refractor before right

Yep!

---

Bors r+

64kramsystem:
> If I'm not mistaken the flop attack staying in place is due to the refractor before right, if so seems ready to go.

The last time I've tested the state PR before it was merged, the problem was the opposite - enemies were always jumping when attacking. Seems to be the reverse, now :grin:

zicklag:
bors cancel

Oh, wait, I didn't know what you guys meant, that's weird.

zicklag:
I think it might be from #221 but I've got to double-check that happens on master right now.

zicklag:
OK yeah, that's happening on master, so it's not from this PR. :)

Bors r+

odecay:
Don't worry it's intended (for now)

odecay:
Also re: flashing box, if I understand what you mean I'm pretty sure it's because when we spawn attacks/collision entities they don't have a Collider shape at first, till they are activated in activate hotbox system which adds the required component.
I think the debug render doesn't deal with the case where there is an entity with a bunch of rapier components but not a Collider shape.

Not 100% certain but it started happening when I changed it to work that way.

zicklag:
Interesting. It's possible for us to filter out certain shapes that rapier gives us to debug render now that we have the custom implementation, so we could potentially filter those out if they aren't actually there.

We can also add a custom debug collor to different colliders with the [`ColliderDebugColor`](https://docs.rs/bevy_rapier2d/latest/bevy_rapier2d/render/struct.ColliderDebugColor.html), so we could use that to debug which colliders are actually getting spawned in the middle of the world.

Another possible cause could be hierarchy flushing issues, maybe? But I think Bevy 0.8 fixed those.

PR #226, opened 2022-08-12
Move attack damage to attack definition in YAML

odecay:
Simple refactor, this is necessary for characters with multiple unique attacks.


odecay:
bors merge

PR #229, opened 2022-08-12
Fix YSort and add editable colliders

edgarssilva:
It fixes y-sorting by using the offset in the `YSort `component (not in use before), also implements hurtboxes and updates hitboxes by using a `ColliderMeta`.

Currently, the hurtbox offset is not implemented since it would require spawning the sensor has a child and I'm not sure if that's necessary, at least for now.

### DIFF
--- a/src/attack.rs
+++ b/src/attack.rs
@@ -65,8 +65,8 @@ fn activate_hitbox(
             {
                 if let Some(fighter_data) = fighter_assets.get(fighter_meta) {
                     commands.entity(entity).insert(Collider::cuboid(
-                        fighter_data.attack.hitbox.x,
-                        fighter_data.attack.hitbox.y,
+                        fighter_data.attack.hitbox.size.x / 2.,
+                        fighter_data.attack.hitbox.size.y / 2.,
                     ));
                 }
             }
--- a/src/fighter.rs
+++ b/src/fighter.rs
@@ -58,8 +58,6 @@ impl Default for Stats {
 }
 
 /// Turns a fighter stub data (loaded from the metadata) into a fully active fighter.
-///
-/// PS: The position of the fighter corresponds the bottom center of the sprite.
 impl ActiveFighterBundle {
     pub fn activate_fighter_stub(
         commands: &mut Commands,
--- a/src/fighter.rs
+++ b/src/fighter.rs
@@ -1,5 +1,4 @@
 use bevy::prelude::*;
-use bevy_rapier2d::prelude::CollisionGroups;
 use rand::prelude::SliceRandom;
 use serde::Deserialize;
 
@@ -59,6 +58,8 @@ impl Default for Stats {
 }
 
 /// Turns a fighter stub data (loaded from the metadata) into a fully active fighter.
+///
+/// PS: The position of the fighter corresponds the bottom center of the sprite.
 impl ActiveFighterBundle {
     pub fn activate_fighter_stub(
         commands: &mut Commands,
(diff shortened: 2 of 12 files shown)


src/attack.rs
```diff
@@ -65,8 +65,8 @@ fn activate_hitbox(
             {
                 if let Some(fighter_data) = fighter_assets.get(fighter_meta) {
                     commands.entity(entity).insert(Collider::cuboid(
-                        fighter_data.attack.hitbox.x,
-                        fighter_data.attack.hitbox.y,
+                        fighter_data.attack.hitbox.size.x / 2.,
```
odecay, line 68: why are you dividing the hitbox by 2 here?

src/fighter.rs
```diff
@@ -59,6 +58,8 @@ impl Default for Stats {
 }
 
 /// Turns a fighter stub data (loaded from the metadata) into a fully active fighter.
+///
+/// PS: The position of the fighter corresponds the bottom center of the sprite.
```
odecay, line 62: :+1:

src/fighter_state.rs
```diff
@@ -679,9 +679,9 @@ fn punching(
                 // Start the attack  from the beginning
                 animation.play(Punching::ANIMATION, false);
 
-                let mut offset = fighter.attack.hitbox_offset;
+                let mut offset = fighter.attack.hitbox.offset;
                 if facing.is_left() {
-                    offset *= -1.0
+                    offset.x *= -1.0
```
odecay, line 684: nice catch

src/fighter_state.rs
```diff
@@ -764,7 +764,7 @@ fn ground_slam(
     {
         // Start the attack
         if let Some(fighter) = fighter_assets.get(meta_handle) {
-            let mut offset = fighter.attack.hitbox_offset;
+            let mut offset = fighter.attack.hitbox.offset;
             if facing.is_left() {
                 offset *= -1.0
```
odecay, line 769: probably apply the same change as above here as well? `offset.x *= -1.0`

src/attack.rs
edgarssilva: The collider expects half widths and heights like in the case of the circle it expects the radius.

src/fighter.rs
edgarssilva: Actually, that is wrong now since I reverted the Anchor.

src/attack.rs
odecay: oh TRUE

odecay: Im now noticing you updated the hitbox sizes to account for it as well so :+1:

odecay:
y_sort_issue
Now we have a different y-sort issue, player can move upward and be standing behind enemies while y-sorted in front.

edgarssilva:
The value needs to be adjusted, should probably make a debug line to see the point.
If it varies too much from asset to asset maybe we change them in the asset file.

edgarssilva:
Been trying to draw a debug line with egui but not sure how to translate the correct coordinates. Any suggestions?

zicklag:
Yeah, it's a little goofy because our egui scale doesn't match the screen scale, it matches the game scale.

Check out the code here where I do the rapier debug rendering:

First we create a central panel without a frame, which means we will get an egui drawable area over the whole screen.

https://github.com/fishfolks/punchy/blob/8e215abbd7615644b364717531b73db65afe3e0f/src/ui/debug_tools.rs#L74-L78

Then we have to use the camera to convert the world coordinates to "normalized device coodinates", so that they will be in the range of ( -1.0, -1.0 ) representing the top left of the screen and ( 1.0, 1.0 ) representing the bottom right.

Then we just have to convert those coordinates to egui points, using the egui size of the current screen ( which can be found with `ui.available_size()` ) and then mutliplying and shifting the NDC coordinates.

https://github.com/fishfolks/punchy/blob/8e215abbd7615644b364717531b73db65afe3e0f/src/ui/debug_tools.rs#L141-L156

edgarssilva:
Haa thanks, I was trying to follow that but didn't create the frame so I had no size.

edgarssilva:
Did just that and nothing showed up :/ The cords don't seem weird to me, I'll have a better look tomorrow.

PR #230, opened 2022-08-13
Debug lines for y-sorting

edgarssilva:
A PR to draw debug lines for the y-sorting that enables easy debugging for fixing current y-sorting problems.

### DIFF
--- a/src/ui/debug_tools.rs
+++ b/src/ui/debug_tools.rs
@@ -1,7 +1,10 @@
 use bevy::prelude::*;
 use bevy_egui::*;
 use bevy_fluent::Localization;
-use bevy_inspector_egui::WorldInspectorParams;
+use bevy_inspector_egui::{
+    egui::{Color32, Stroke},
+    WorldInspectorParams,
+};
 use bevy_rapier2d::{
     plugin::RapierContext,
     prelude::{ColliderDebugColor, DebugRenderContext},
@@ -11,7 +14,7 @@ use bevy_rapier2d::{
     },
 };
 
-use crate::localization::LocalizationExt;
+use crate::{camera::YSort, localization::LocalizationExt};
 
 /// System that renders the debug tools window which can be toggled by pressing F12
 pub fn debug_tools_window(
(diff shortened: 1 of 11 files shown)


erlend-sh:
This is a visual thing, right? Could you please share a screenshot of the change in action?

edgarssilva:
> This is a visual thing, right? Could you please share a screenshot of the change in action?

It's still in progress, it's just basically a line at the bottom of each sprite representing the point at which sprites appear behind each other. Just to help debug that value.

When It's ready I can get a screenshot.

src/ui/debug_tools.rs
```diff
@@ -163,3 +178,66 @@ impl<'world, 'state, 'a, 'b, 'c> DebugRenderBackend
...
+        for (ysort, transform) in query.iter() {
+            egui::CentralPanel::default()
+                .frame(egui::Frame::none())
+                .show(egui_context.ctx_mut(), |ui| {
+                    let mut a = transform.translation.clone();
+                    a.x += 16.;
+                    a.z = 0.;
+
+                    let mut b = transform.translation.clone();
+                    b.x += 16.;
```
zicklag, line 221:
```suggestion
                    b.x -= 16.;
```

Hey @edgarssilva this is why the line wasn't displaying. Point `a` and `b` were the same point, so a line between them was invisible.

I found out by rendering a circle at each point instead of a line between them.

edgarssilva: Oh god I'm blind, thanks for picking it up.

zicklag: No problem, I've been there before. :D

edgarssilva:
So after adding the debug lines it was easy to tell that the problem is caused because of the different padding between sprite sheets. I think adding a `fighter_size` property to the fighters is a good idea and can be beneficial later on.

edgarssilva:
Ok so I just finished up everything should be working fine now. 
One thing we should be careful of is that skins of the same fighter should be packed into a sheet the same way.
Right now we have one case where this happens but it's a very small one that won't be noticeable. But to keep in mind in the future.

different_padding

zicklag (approved):
Haven't tested but the code looks good. :+1:

edgarssilva:
@erlend-sh Here you go.

image

edgarssilva:
bors+

Issue #84, opened 2022-07-14
Breakable boxes which can contain other items


zicklag:

Zac8668:
I'll claim this, I have some idea of how to do it, gonna follow how the bottle and health were made.

Zac8668:
I'm not sure how the boxes would function in-game, here's some options that I thought:

Throwable boxes that break on the ground, (maybe enemies or players) that drop the item where it broke.
Simple item that just drops the item on the ground when used, like the health.
Box that breaks when a enemy and/or player hits it, and drops the item where it was.

Which route should I take?

odecay:
The final design for these items would be something like this.
A crate, entity spawned with `Damageable` and `Health` components.
It should be able to contain another item.
It should be able to be broken by attacks after a certain amount of damage is taken.
It should be able to be picked up and carried above the head of the player, preventing them from performing other attacks while held.
It should be able to be thrown, when it hits something it should deal damage and knockback, and it should then break, exposing its contained item.

I think to start with, this issue should just tackle the first bits of that. A crate which can contain an item and be damaged by attacks, when its health is depleted it should break and allow the player to access the item inside.

Zac8668:
@odecay I think this can be closed

zicklag:
I agree, we can open specific issues for future improvements. :+1:

PR #233, opened 2022-08-17
Add projectile despawning on collision.

### LINKED ISSUE #210: Make Projectiles Despawn on Collision Again
### Description

After #203 is merged we need to make sure projectiles despawn on collisions again, because that functionality got lost in the refactor.

### Alternatives & Prior Art

_No response_

edgarssilva:
Added a new `DespawnOnCollision` component that takes a number of collisions before it is respawned on collision. Added this component to the projectile bundle with the value of 0.

I do not really like the naming on this one, open to ideas :), should fix #210.

### DIFF
--- a/src/attack.rs
+++ b/src/attack.rs
@@ -24,7 +24,7 @@ impl Plugin for AttackPlugin {
                     .run_in_state(GameState::InGame)
                     .with_system(activate_hitbox)
                     .with_system(deactivate_hitbox)
-                    .with_system(despawn_collision)
+                    .with_system(breakable_system)
                     .into(),
             )
             // Attack damage is run in PostUpdate to make sure it runs after rapier generates collision events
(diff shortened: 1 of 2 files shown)


zicklag (approved):
Looks good!

zicklag:
As far as name, maybe we make it `Breakable`, and then add a field in it for the `hits` or something like that? I've not got any other much better ideas. That one's not easy to name.

odecay:
> As far as name, maybe we make it `Breakable`, and then add a field in it for the `hits` or something like that? I've not got any other much better ideas. That one's not easy to name. man_shrugging

I sort of agree with this but I think if we do that we should consider tying projectile breakability in with item breakability system in #84.

edgarssilva:
> I sort of agree with this but I think if we do that we should consider tying projectile breakability in with item breakability system in #84.

For that we could create a `Drop` component that drops an item when breaking or other conditions like on despawn.

src/attack.rs
```diff
@@ -40,6 +41,19 @@ pub struct Attack {
     pub velocity: Vec2,
 }
 
+/// A component that depawns an entity after collision with tolerance to x collisions.
+#[derive(Component, Clone, Copy, Default, Reflect)]
+pub struct Breakable {
+    pub hits: i32,
+    count: i32,
```
zicklag, line 48: My gut reaction is that this should be public. Seems like we might as well allow other systems to modify it it sees fit ( maybe mods for example ). It doesn't really hurt anything other than maybe producing unexpected behavior, which if somebody is modifying that field manually, then it's kind of already obvious what they're messing with.

I you feel like we should keep it private, though, I'm not super set on making it public.

If we do make it public, maybe name the fields, `hit_tolerance` and `hit_count`, and add super minimal doc strings on them to help differentiate what they are for?

zicklag (approved):
Awesome, looks great!

If you're ready feel free to have bors merge it.

PR #236, opened 2022-08-21
button sounds

DRuppFv:
Note: I removed a lot of code that was not done

Just need help with playing the audio, the audio files are in the PR


DRuppFv:
I believe it is done.
I just need someone to test it (my computer is not making sounds).

64kramsystem:
They sound ok to me. I would (personal taste) increase the volume a bit, but it's fine as now.

When I click on "Start Game", I don't hear anything; I don't know if this is intended or not (also on Quit, but that's more likely to be intended).

64kramsystem (commented):
Tested; notes reported in the PR discussion.

DRuppFv:
> They sound ok to me. I would (personal taste) increase the volume a bit, but it's fine as now.
> 
> When I click on "Start Game", I don't hear anything; I don't know if this is intended or not (also on Quit, but that's more likely to be intended).

The quit thing can be easily fixed. Just don't close the game immediately. About the start button, it should just play a 
different sound, and i don't know the reason. For now, i'll try to fix my computer before updating this PR

DRuppFv:
@64kramsystem 
I fixed my computer. The "start game" plays the sound! It's just a different one, if you don't like it, i can change.
About the quit game button... I don't think would be good just make the program wait for 0.5 seconds before it close, then, the best option would be playing the audio in the own "quit game" button code. What do you think about it?

64kramsystem:
> I fixed my computer. The "start game" plays the sound! It's just a different one, if you don't like it, i can change.

Ah! I've noticed it. I'm neutral to whatever is used, however, it's hard to notice.

DRuppFv:
> > I fixed my computer. The "start game" plays the sound! It's just a different one, if you don't like it, i can change.
> 
> Ah! I've noticed it. I'm neutral to whatever is used, however, it's hard to notice.

It's a "normal" sound of "Start Game" button, you can see something like this in a lot of games like Punchy, but you are right. It would be better if we had an loading screen.

And i just put a new one btw

odecay (approved):

src/audio.rs
```diff
@@ -92,6 +93,26 @@ pub fn animation_audio_playback(
...
 }
 
+/// Plays main menu sounds
+pub fn main_menu_sounds(
+    mut context: ResMut<EguiContext>,
+    effects_channel: Res<AudioChannel<EffectsChannel>>,
+    asset_server: Res<AssetServer>,
+) {
+    for event in &context.ctx_mut().output().events {
+        println!("{:?}", event);
```
odecay, line 103: This should be removed

odecay (changes requested):
It seems like this is almost ready, could you clean up the println?

odecay (approved):

odecay:
bors merge

PR #238, opened 2022-08-22
Hurtbox size update

odecay:
Updated fighter hurtboxes to be more reflective of their sprites.
This was particularly necessary for the boss, which was previously very difficult to damage.

This also adds some assets for a dev fighter for later use.


odecay:
bors merge

PR #240, opened 2022-08-23
Breakable Box

Zac8668:
Breakable box that contain other items.
#84


Zac8668:
The functionality of the box is done, breaking, dropping and file loading the inside item, but for some reason the item dropped image does not show.

Zac8668:
PR is done aside from the dropped item image not loading.

src/attack.rs
```diff
@@ -62,6 +65,15 @@ impl Breakable {
...
 }
 
+/// A component that with Breakable, drops a item when broke.
+#[derive(Component)]
+pub struct Drop {
+    /// Item data
+    pub item: ItemMeta,
+    /// Set true to drop
+    pub drop: bool,
+}
```
odecay, line 75: I think this may fit better in `item.rs`

src/attack.rs
```diff
@@ -169,3 +183,25 @@ fn breakable_system(
...
+                item: String::new(),
+                item_handle: items_assets.add(drop.item.clone()),
+            };
+            let item_commands = commands.spawn_bundle(ItemBundle::new(&item_spawn_meta));
+            ItemBundle::spawn(item_commands, &item_spawn_meta, &mut items_assets);
+
+            commands.entity(entity).despawn_recursive();
+        }
+    }
+}
```
odecay, line 207: Maybe this could be moved to `item.rs` as well, but it would require either implementing plugin for `ItemPlugin`, or adding the system in main.rs.

src/item.rs
```diff
@@ -31,13 +31,38 @@ impl ItemBundle {
...
+        if let Some(item_meta) = items_assets.get_mut(&item_spawn_meta.item_handle) {
+            if let ItemKind::BreakableBox {
+                hurtbox,
+                hits,
+                item,
+                ..
+            } = &item_meta.kind
+            {
+                let mut physics_bundle = PhysicsBundle::new(hurtbox, BodyLayers::ENEMY);
+                physics_bundle.collision_groups.filters = BodyLayers::PLAYER_ATTACK;
```
odecay, line 55: we should probably add a new bodylayer in `impl BodyLayers` in collisions.rs  `pub const BREAKABLE_ITEM: u32 = 1 << 4;` and use that here as the membership, keeping the filter as `BodyLayers::PLAYER_ATTACK` as you have it.

src/fighter_state.rs
```diff
@@ -955,9 +963,19 @@ fn dying(
...
+        (
+            Entity,
+            &Transform,
+            &Facing,
+            &mut Inventory,
+            Option<&Children>,
+        ),
+        With<Throwing>,
+    >,
+    being_hold: Query<Entity, With<BeingHold>>,
```
odecay, line 976: I feel it might make more sense to add an entirely new player state, something like `holding` and its transition system `transition_from_holding` with a higher priority than attacking, which I think is generally how we intend to structure our state model going forward, and would remove the need for checking if we are holding in the attack system as well since the higher priority state would simply take precedent when the transitions were determined.

odecay (changes requested):
Soft request of changes here, I'd like to hear feedback from other contributors as well

src/attack.rs
edgarssilva: Imo I would prefer for breakable to throw an event and this system just listens to it

assets/items/box/box.item.yaml
```diff
@@ -0,0 +1,25 @@
...
+    hurtbox:
+        size: [36, 52]
+        offset: [0, 0]
+
+    item:
+        name: Bottle
+
+        image:
+            image: bottle.png
+            image_size: [11, 31]
```
edgarssilva, line 21: I'm not sure how this gets implemented but we could import the item yaml file here as we do with fighters in the level. @zicklag might know a little better about this.

src/attack.rs
```diff
@@ -152,15 +164,17 @@ fn attack_damage_system(
...
 ) {
     for ev in events.iter() {
         if let CollisionEvent::Started(e1, e2, _flags) = ev {
             for e in [e1, e2].iter() {
-                if let Ok(mut breakable) = despawn_query.get_mut(**e) {
+                if let Ok((mut breakable, drop)) = despawn_query.get_mut(**e) {
                     if breakable.hit_count < breakable.hit_tolerance {
                         breakable.hit_count += 1;
+                    } else if let Some(mut drop) = drop {
+                        drop.drop = true;
```
edgarssilva, line 177: As I said I would think of throwing an event here but not sure how the others feel about that.

assets/items/box/box.item.yaml
zicklag: Yes, I think we should specify the path to the item here, and then in item asset loader we'd use code similar to the code here, to load the item handles from the path:

https://github.com/fishfolks/punchy/blob/d8bfbdcda79e576025c1e0367307a601844a122d/src/assets.rs#L191-L199

src/attack.rs
zicklag: An event makes sense to me here.

We use a similar pattern with the attack collision system, so at least there's some precedent, too. The system responsible for detecting a certain kind of collisions emits an event when one occurs so other systems can supplement the behavior.

src/attack.rs
zicklag: As mentioned in my other comment, :+1: for using an event, but we still need the system, so I think it would make sense to add to the `ItemPlugin`.

src/lifetime.rs
```diff
@@ -15,14 +17,18 @@ pub struct Lifetime(pub Timer);
...
+    mut entities: Query<(Entity, &mut Lifetime, Option<&mut Drop>)>,
     time: Res<Time>,
 ) {
-    for (entity, mut lifetime) in &mut entities {
+    for (entity, mut lifetime, drop) in &mut entities {
         lifetime.tick(time.delta());
 
         if lifetime.finished() {
-            commands.entity(entity).despawn_recursive();
+            if let Some(mut drop) = drop {
```
zicklag, line 27: I don't like having to make the lifetime system specifically tailored for the droppable items.

I think this is another argument in favor of using an event to drop items.

Maybe we make a `LifetimeExpired` event, that will be emitted by the `lifetime_system` when something is deleted by it. This keeps the lifetime plugin fully agnostic of the rest of the plugins.

Then we can have the `drop_system` listen for `LifetimeExpired`, or `ItemDropped` ( or whatever it is called ) events.

src/item.rs
```diff
@@ -31,13 +31,38 @@ impl ItemBundle {
...
+        items_assets: &mut ResMut<Assets<ItemMeta>>,
+    ) {
         let ground_offset = Vec3::new(0.0, consts::GROUND_Y, consts::ITEM_LAYER);
         let transform_bundle = TransformBundle::from_transform(Transform::from_translation(
             item_spawn_meta.location + ground_offset,
         ));
 
         commands.insert_bundle(transform_bundle);
+
+        if let Some(item_meta) = items_assets.get_mut(&item_spawn_meta.item_handle) {
```
zicklag, line 46: This should actually be a panic if the asset isn't there, because this logic _needs_ to run for the item to spawn correctly.

We are already pre-loading all the assets before spawning the level, so the asset get shouldn't fail unless there is a bug we need to fix with the pre-loader, and we'll want to know about it in that case so panicking is still the right approach I think.

zicklag (changes requested):
I agree the points made by @odecay and @edgarssilva, left my :+1: additional comments.

src/lifetime.rs
Zac8668: I added events, made the `LifetimeExpired`, but I added a `Drop` and `Transform` fields to it, not sure how to reference the new code on `lifetime.rs` but could you take a look and see if it's agnostic enough? Tried making the event with just a entity field but found some complications to get the `Drop` and `Transform` on the `drop_system` with just a entity.

Zac8668:
Sorry for the review request mess, didn't knew I only could request one

src/fighter_state.rs
Zac8668: I implemented the states but not sure if it's how you wanted, or if it's a good implementation.

src/fighter_state.rs
```diff
@@ -274,6 +285,9 @@ impl Dying {
     pub const ANIMATION: &'static str = "dying";
 }
 
+#[derive(Component)]
+pub struct BeingHold;
```
zicklag, line 289: I think maybe this should be:

```suggestion
pub struct BeingHeld;
```

src/fighter_state.rs
```diff
@@ -592,7 +607,7 @@ fn flopping(
                             BodyLayers::ENEMY_ATTACK
                         },
                         if is_player {
-                            BodyLayers::ENEMY
+                            BodyLayers::ENEMY + BodyLayers::BREAKABLE_ITEM
```
zicklag, line 610: I think this needs to be like this:

```suggestion
                            BodyLayers::ENEMY | BodyLayers::BREAKABLE_ITEM
```

src/fighter_state.rs
```diff
@@ -707,7 +722,7 @@ fn punching(
...
```
zicklag, line 725:
```suggestion
                            BodyLayers::ENEMY | BodyLayers::BREAKABLE_ITEM
```

src/item.rs
```diff
@@ -89,9 +129,50 @@ impl Projectile {
             collision_types: ActiveCollisionTypes::default() | ActiveCollisionTypes::STATIC_STATIC,
             //TODO: define collision layer based on the fighter shooting projectile, load for asset
             //files of fighter which "team" they are on
-            collision_groups: CollisionGroups::new(BodyLayers::PLAYER_ATTACK, BodyLayers::ENEMY),
+            collision_groups: CollisionGroups::new(
+                BodyLayers::PLAYER_ATTACK,
+                BodyLayers::ENEMY + BodyLayers::BREAKABLE_ITEM,
```
zicklag, line 134:
```suggestion
                BodyLayers::ENEMY | BodyLayers::BREAKABLE_ITEM,
```

zicklag (changes requested):
I haven't tested, but I did a quick review and I think this looks good!

I just had a tiny suggestion on a name, and I'm fairly sure collision layers must be combined with the "bitwise or" operator ( `|` ) not the `+` operator. Good job!

zicklag (approved):

odecay (approved):
Code looks really good!
Going to quickly test then merge :+1:

odecay:
bors merge

PR #247, opened 2022-08-30
Change Y-Sort value to represent the base Z value

edgarssilva:
Basically, the Y-Sort should take a value with the default Z and maybe some offset if necessary.
Currently, the default Z is defined as a constant with a value of 300.

### DIFF
--- a/src/camera.rs
+++ b/src/camera.rs
@@ -25,8 +25,8 @@ impl Plugin for CameraPlugin {
 }
 
 /// Component to sort entities by their y position.
-/// Takes in a offset usually the sprite's height.
-/// the offset is currently unused, but may be used again if we implement virtual z for jumping
+/// Takes in a base value usually the sprite default Z with possibly an height offset.
+/// this value could be tweaked to implement virtual Z for jumping
 #[derive(Component, Default, Reflect)]
 #[reflect(Component)]
 pub struct YSort(pub f32);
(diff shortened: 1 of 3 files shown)


odecay:
I think I agree with this yup.
I cant remember currently do items have a Ysort?
If so they should also be initialized with that offset,
if not i think they probably should have it, but we may actually have to apply more complex sorting since I didnt address them at all when making changes to the anchor of character sprites.

edgarssilva:
Currently, they don't but I can add it up.

odecay:
we might have to change the anchor for item sprites the same way I did for character sprites, although it could also cause problems when doing rotations like in the bottle because rotations occur relative to the anchor point.

I did have another idea for a way to implement sorting and position that might make more sense than what I already did.
I was thinking we could have some sort of position component, maybe a newtype around transform, or just a child entity with a transform, that represented the feet of a character, or bottom of an item, and then refactor all movement bound checking and positional placement to be based off that component rather than the main transform. Conceptually it would be similar to what I already implemented where we think in position at base of feet, but it might be more flexible/require less offsetting other places, like for the hitboxes/hurtboxes.

edgarssilva:
Yeah, that looks nice maybe we could work everything on it and then get the currect transform from it before rendering.

edgarssilva:
I'll merge this one for now and leave the new position to a new issue.

Issue #86, opened 2022-07-14
Melee Weapons


zicklag:
Melee weapons are items which while held change the fighters available attacks.

My proposed design for this would be to move available attacks to a component with a hashmap/vec of attacks which is attached to each player.
Items with attacks should have their attacks added to `AvailableAttacks` component of the fighter they are picked up by while held, with the ability to take priority over the fighters innate attacks (dependent on design of weapon).

For non weapon items, we could decide to either block the fighter from using a subset of their attacks while holding them, through the same method or default to "the only attack you can do while holding a non weapon item is to throw the item, freeing up your hands to continue attacking."

This also calls into question whether all items should be required to have an attack, and/or if we should change the damage field in item definitions to be throw damage instead (since damage stat in fighters has now moved to being attack specific)
This also raises the question of how best to represent the animations of weapon attacks, since they will implicate both the fighters spritesheet and weapon spritesheet.

Zac8668:
I'll claim this

PR #254, opened 2022-09-14
Melee Weapon

Zac8668:
Add Melee Weapon
#86


zicklag (approved):
This looks great!

The only thing I notice is that we might want to tweak a weapon offset of some sort for the positioning relative to the player.

image

Still, I don't think that needs to block this PR, and we can do that after this is merged.

Zac8668:
> The only thing I notice is that we might want to tweak a weapon offset of some sort for the positioning relative to the player.
> 
> image

@zicklag maybe this could be a field on the yaml file? Cause there's a variety of weapons that could be implemented, and with that we could have a negative offset, something that a player relative offset would not probably not allow, and as I noticed it now, the sword on jumpy has a negative one: 
image

zicklag:
Yeah, that's exactly what I was thinking. :+1:

src/metadata.rs
```diff
@@ -136,6 +136,11 @@ pub enum ItemKind {
         #[serde(skip)]
         item_handle: Handle<ItemMeta>,
     },
+    MeleeWeapon {
+        attack: AttackMeta,
+        audio: AudioMeta,
+        spritesheet: Box<FighterSpritesheetMeta>,
```
odecay, line 142: Is there a reason we need to box here? (not saying its wrong, im just not super acquainted with Boxing)

src/player.rs
```diff
@@ -59,6 +67,10 @@ impl PlayerBundle {
             fighter_handle,
             input_manager_bundle,
             inventory: Inventory(None),
+            available_attacks,
         }
     }
 }
+
+#[derive(Component)]
+pub struct AvailableAttacks(pub Vec<AttackMeta>);
```
odecay, line 76: :+1:

odecay (commented):
Looks good.

src/metadata.rs
Zac8668: Clippy said it's best because of the size difference between the ItemKind Variants, I was surprised too

zicklag: Yeah, because an enum must take up as much memory as it's largest variant, and the `FighterSpritesheetMeta` struct makes the `MeleeWeapon` variant much larger than all the other variants, boxing it can make the enum take less space when using the non-`MeleeWeapon` variants.

I'm not sure if it's really a concern, but seems like we might as well trust clippy in this case.

zicklag:
Awesome, thanks a bunch!

bors merge

PR #255, opened 2022-09-18
Scripting MVP

### LINKED ISSUE #253: Create Scripted In-Game Item
### Description

Now that `bevy_mod_js` scripting has browser support, we try it out for scripting in Punchy!

The objective is to make _something_ in the game that uses scripting for it's logic. This could be an attack, or an item.

It will require some experimentation to determine exactly how to architect it and whether we should start with an item or an attack.

### Alternatives & Prior Art

_No response_

zicklag:
Closes #253.

### DIFF
--- a/lib.bevy.types.ts
+++ b/lib.bevy.types.ts
@@ -1,4 +1,5 @@
 // generated with https://github.com/jakobhellermann/bevy_reflect_ts_type_export
+// bevy_core
 type bool = boolean;
 type f32 = number;
 type f64 = number;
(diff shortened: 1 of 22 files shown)


zicklag:
OK, this is ready for review!

We want to wait for https://github.com/jakobhellermann/bevy_mod_js_scripting/pull/23 before merging, but we should have that done within a day.

---

This PR removes the `Health` item kind and adds a new `Script` item kind. Then we use a script to re-implement the health item.

Additionally, you can add scripts to the game without tying them to an item. I added a nifty [demo script](https://github.com/fishfolks/punchy/blob/c410b9db925c359b5e26a5f3016eeb1d8ceb4a4f/assets/scripts/demo_script.ts) to show how that can be used, and acts as a short tutorial to the basics of the scripting system.

64kramsystem (approved):

zicklag:
I just updated to use the latest master version of `bevy_mod_js_scripting` and this is ready to merge!

It's a big update, though, so I'll wait for @odecay to give at least a casual approval.

A quick summary of what's included:

- Added a `.editorconfig` file. This just helps VSCode play nicer with the indent size difference between Rust and JS depending on the formatters you have installed.
- Updated the version of wasm_bindgen
- Allowed you to specify a list of scripts in the `.game.yaml` file that not specifically tied to any specific item or attack.
  - Added a demo script of this kind
- Added events that are emitted when a script item is grabbed or thrown. This is used in scripts to carry out their logic when their item is interacted with.
  - Many items still cannot be implemented with scripts. This is work-in-progress so we still have non-scripted items. 
- Changed the health item type to a script item type and re-implemented it with a script.
- Added typescript definitions for the scripting API and a TypeScript config file to enable auto-completion and type checking for scripts in VSCode.

Also, all scripts can be hot reloaded, so I'm literally able to modify the game while it's running, which is quite fun!

That's going to be serious motivation to start moving more gameplay design elements such as attacks and other items to scripts so we can hot reload while designing and tweaking the gameplay.

lib.bevy.types.ts
```diff
@@ -0,0 +1,613 @@
+// generated with https://github.com/jakobhellermann/bevy_reflect_ts_type_export
```
odecay, line 1: im guessing this will need to be updated with every bevy version bump?

src/scripting.rs
```diff
@@ -0,0 +1,15 @@
...
+
+mod ops;
+
+pub struct ScriptingPlugin;
+
+impl Plugin for ScriptingPlugin {
+    fn build(&self, app: &mut App) {
+        let custom_ops = ops::get_ops();
+
+        app.insert_non_send_resource(JsRuntimeConfig { custom_ops })
```
odecay, line 12: What does it mean that its non send? Only accessible from main thread?

odecay (approved):
I went over the code, I dont understand the ops.rs file. The api for scripting as I understand it seems pretty nice though. I was expecting to see some system that dealt with the events that were being registered like 'ScriptItemThrowEvent`. I see they are fired in `fighter_state`, am I missing another system that deals with those events on the rust side or is that all thats needed to run the js/ts?

lib.bevy.types.ts
zicklag: For now, but once Bevy 0.9 is released we can auto-generate the file from a `build.rs` script so there will be no manual effort required to sync.

src/scripting.rs
zicklag: Yes. Since `JsRuntimeConfig` contains structs that may not implement `Send`, bevy has to make sure it only gets inserted and accessed on the main thread.

zicklag:
Maybe I overdid it, but I pushed a commit with a ton of comments to `ops.rs` that hopefully helps explain the code.

There are a couple things in there, specifically around `JsValueRefs` that kind of requires knowing how `bevy_mod_js_scripting` works to fully understand, and it's not really documented yet because it's brand new. But hopefully the comments give a general overview of what's happening.

> I was expecting to see some system that dealt with the events that were being registered like 'ScriptItemThrowEvent. I see they are fired in fighter_state`, am I missing another system that deals with those events on the rust side or is that all thats needed to run the js/ts?

That's what the `ItemGetGrabEvents` op does in `ops.rs`. Because the out-of-the-box scripting API from `bevy_mod_js_scripting` doesn't have a way to read Bevy events, we have to add a custom `JsRuntimeOp` that adds a `punchy.getItemGrabEvents()` function to the JavaScript environment.

This function will read all the `ScriptItemGrabEvent`s sent by the code in `fighter_state`.

Later, we will also need to add another JS op for getting the throw events ( or maybe we should merge both events into the same struct so that we don't need two different kinds of events? :thinking:  ).

---

For anything that the out-of-the-box scripting API doesn't do yet, we can just create our own custom JS ops that can do anything we want, so it's super easy to add new functionality to the scripting interface.

erlend-sh:
Does the [scripting wiki](https://github.com/fishfolks/punchy/wiki/Scripting) need any updates?

And is this still valid?
> Since the scripting API is in flux, we won't include a tutorial here

zicklag:
> Does the [scripting wiki](https://github.com/fishfolks/punchy/wiki/Scripting) need any updates?

I wrote it as if this was already merged so it's up-to-date.

> And is this still valid?

Yeah, the demo script I linked to will exist once this is merged and that is heavily commented, acting somewhat like a tutorial.

Now that you mention it, though, I don't like the wording very much. I'll just refer to the demo script _as_ the tutorial. That will also help make sure it stays up-to-date because the game will error when we run it if something is wrong in that file.

erlend-sh:
I updated the wiki with your [Deno explanations](https://github.com/fishfolks/jumpy/discussions/468) over in Jumpy-land 👌 

LGTM?

zicklag:
Looks great, and good idea, thanks!

odecay:
bors merge

zicklag:
bors retry

zicklag:
bors retry

PR #259, opened 2022-09-27
making button sounds random + new sounds

DRuppFv:
Now the sounds are random.


zicklag:
Thanks! I'll try to review before the end of the day.

odecay (approved):

src/audio.rs
```diff
@@ -106,7 +107,10 @@ pub fn main_menu_sounds(
                 effects_channel.play(asset_server.load("ui/down_play_button.ogg"));
             } else {
                 //Play one of the down button audios, except down_play_button
-                effects_channel.play(asset_server.load("ui/down_button_1.ogg"));
+                effects_channel.play(asset_server.load(&format!(
```
zicklag, line 110: Running `asset_server.load()` will unfortunately cause Bevy, at least on WASM, to re-download the asset from the network, so we will want to load all of our sound handles once, and then pick a random handle here.

That said, if we just add a TODO or an issue to come and update it later, that could be fine for now, and we could just merge it.

DRuppFv: What do you think could be done to fix that?

zicklag: We would probably do something similar to what we did [here](https://github.com/fishfolks/punchy/blob/1e94c180cbd244608775ac134ff5088693d82b5c/src/assets.rs#L272) for the random player images, where we load a list of sounds from the `.game.yaml` file, load them, and then create a list of handles for each sound.

Then we can choose a random sound handle out of that list in this `main_menu_sounds` system.

It's a little bit more involved so I might be able to get to that tomorrow if you wanted.

zicklag: I pushed an update that adds the sounds to the game metadata, so we're good to merge now. :rocket:

PR #260, opened 2022-09-29
Aesthetic attachments (hats++)

### LINKED ISSUE #77: Aesthetic attachments (hats++)
There is a reference implementation of this in Jumpy:
- https://github.com/fishfight/FishFight/pull/381

Though in Punchy it gets more advanced because our characters have more varied movement sets that need to be animated properly. A hat attachment is harder to do right if a character can perform a head-butt, tilting their head at an angle.

CE1C9DB6-DDCB-4727-AFE9-DCFFAA818E48
99E6F920-0854-4814-801B-9AA5532022C8
AF466C6A-8A07-43A7-BAEE-2A182F73B344

Even harder if a character has a tattoo variation:
4DABF2ED-3685-43D5-B25C-FC0B3DFE02AF
0F0AC534-1310-4565-ACA2-7062CE9172B7

Yet we’re obviously looking at the same core model for both of these examples respectively. What techniques are available to us if we wanna treat these aesthetic variants as *attachments* that don’t require us to fully hand-paint each new character variant?

Zac8668:
This will add the aesthetic attachments, but also a attachment component to deal with attaching things to the fighters.
Closes #77


Zac8668:
I think this is completed, the thing we could add is a folder called `assets/attachmets` or `assets/hats` and have some hats pre configured there, so when adding we could reference it, as we do with items on `...level.yaml`, not sure if that's necessary.

odecay:
Ok first comment, why does this change the dimensions of the sharky sprite?

Zac8668:
> Ok first comment, why does this change the dimensions of the sharky sprite?

There isn't enough space for the hat on the current sprite

src/fighter.rs
```diff
@@ -124,9 +132,88 @@ impl ActiveFighterBundle {
...
+}
+
+#[derive(Component)]
+pub struct Attached {
+    /// Syncs facing with parent facing
+    pub sync_facing: bool,
+    // Syncs animation with parent animation
+    pub sync_animation: bool,
+    /// Change position based on facing
+    pub position_face: bool,
```
odecay, line 179: Another thing we could consider is having the actual attach point be child component of the fighter, and which gets translated in x when the fighters facing changes. Then whatever is attached to that attach point would be a child entity of that and get moved automatically.

src/fighter.rs
```diff
@@ -17,6 +17,14 @@ use crate::{
     player::Player,
 };
 
+pub struct FighterPlugin;
+
+impl Plugin for FighterPlugin {
+    fn build(&self, app: &mut App) {
+        app.add_system_to_stage(CoreStage::PostUpdate, attachment_system);
```
odecay, line 24: Noting that you might want to put this in `CoreStage::Update`, but I could be wrong, will return for a closer look.

src/metadata.rs
```diff
@@ -109,6 +109,7 @@ pub struct FighterMeta {
     pub hurtbox: ColliderMeta,
     //Will likely need a hashmap(?) of AttackMetas, fighters will have multiple attacks
     pub attack: AttackMeta,
+    pub attachment: Option<FighterSpritesheetMeta>,
```
odecay, line 112: maybe could decide on another way to represent spritesheets of types other than `FighterSpritesheet`

odecay (commented):
Did a first pass, looking good so far, Ill come back to it for another look tomorrow.
Syncing strategy changes definitely seemed like they helped clean things up.

Im wondering if the spritesheet cloning is the best way to go about making the spritesheet for the hat, but I can't say for sure. 
I think it makes sense given we currently are using a spritesheet for the hat that matches the player spritesheet, but I wonder what would happen if an attachment was added to a non player fighter?

src/fighter_state.rs
```diff
@@ -1433,23 +1443,19 @@ fn holding(
...
+        (
+            Entity,
+            Option<&mut MeleeAttacking>,
+            Option<&Player>,
+            Option<&Enemy>,
+            &AvailableAttacks,
+            &mut LinearVelocity,
+            &Facing,
+        ),
+        Without<MeleeWeapon>,
```
zicklag, line 1456: Is this `Without` filter necessary?

Same for the `Without<ProjectileWeapon>` below.

src/fighter.rs
zicklag: All the state handler functions run in `Update`, so I think `PostUpdate` is right.

zicklag (commented):
Just one question, but otherwise, it looks good to me! :+1:

src/fighter_state.rs
Zac8668: Just checked and they're not, they were before, but with the Attached, not anymore.

src/fighter.rs
Zac8668: This actually solves a problem when grabbing a item with the player facing left, the attached item appears facing right for one frame and then facing left, this is because the facing value for the item is first set to default/right, and just when the system is run it fixes it.

odecay:
Okay I tested this out. There is one issue with sprite sorting the hat attachment relative to the sword, hat overlays the sword.
hat_sort_punchy

odecay:
Okay looks good

odecay:
bors merge

PR #266, opened 2022-11-04
Remove unused files

DRuppFv:
I added these files when I was working with button sounds, but they will no longer be used.


odecay:
bors merge

PR #270, opened 2022-11-08
Beginning support for multiple attacks per fighter.

odecay:
~This broke AI as of now, I must have missed some implications of the change, will take another look shortly.
Probably the using the last `AvailableAttack` can have already changed when we are checking its properties before attacks finish.~

Edit: Fixed, should be good to go.


odecay:
bors merge

PR #272, opened 2022-11-10
Boss Bomb Throwing

### LINKED ISSUE #269: Implement boss bomb throwing
Boss bomb attack.
Attack should throw projectile bombs in an arc which land on the ground some distance away, after a short delay or upon collision they should explode causing AOE damage and Knockback state to players in radius.

- [ ] Find bomb spritesheet from Jumpy
- [ ] Implement state for bosses bomb throw attack
- [ ] Add attack to boss YAML

Check bottle throw for projectile reference.

Zac8668:
Implements boss bomb throwing
Closes #269


src/fighter_state.rs
```diff
@@ -1011,6 +1054,60 @@ fn ground_slam(
...
+        &mut fighters
+    {
+        // Start the attack
+        if let Some(fighter) = fighter_assets.get(meta_handle) {
+            let mut offset = fighter.attack.hitbox.offset;
+            if facing.is_left() {
+                offset.x *= -1.0
+            }
+            offset.y += fighter.collision_offset;
+            let attack_frames = fighter.attack.frames;
```
odecay, line 1082: Instead of using the attack.frames metadata now we are using the attack.frames from the last attack on `AvailableAttacks`

odecay (changes requested):

src/metadata.rs
```diff
@@ -121,6 +121,8 @@ pub struct AttackMeta {
     pub hitbox: ColliderMeta,
     pub hitstun_duration: f32,
     pub velocity: Option<Vec2>,
+    pub spritesheet: Option<FighterSpritesheetMeta>,
+    pub projectile_frames: Option<AttackFrames>,
```
odecay, line 125: Could you refactor this this to remove the optional spritesheet from `AttackMeta` and move that to an item definition?

odecay (changes requested):

src/metadata.rs
Zac8668: For sure, just not sure how would the boss get the item, a field on the fighter? or maybe on the attack?

odecay: You dont have to have it pick up or hold the item or anything, just spawn it when the boss attacks

src/fighter_state.rs
```diff
@@ -1218,6 +1365,9 @@ fn throwing(
                         }
                     }
                 }
+                ItemKind::Bomb { .. } => {
+                    panic!("Can't throw bomb")
+                }
```
odecay, line 1370: It might be interesting to come back in future and make it so the player can pickup and throw the bomb before it explodes.

odecay (commented):
Ok Its looking pretty good I think, might do another pass tomorrow but :+1:

src/assets.rs
```diff
@@ -308,13 +319,14 @@ impl AssetLoader for FighterLoader {
...
 
+            let index_len = meta.spritesheet.image.len();
             if let Some(ref mut attachment) = meta.attachment {
                 for (index, image) in attachment.image.iter().enumerate() {
                     let (texture_path, texture_handle) =
                         get_relative_asset(load_context, load_context.path(), image);
 
                     let atlas_handle = load_context.set_labeled_asset(
-                        format!("atlas_{}", index + attachment.image.len()).as_str(),
+                        format!("atlas_{}", index + index_len).as_str(),
```
odecay, line 329: whats the reason for change here?

Zac8668: This was done when the attack had a spritesheet, there's no reason for it now, gonna remove it

src/fighter_state.rs
Zac8668: For sure, I thing the only problem is that the bomb stays "fusing" for I think 0.36 seconds, so not much time to grab and throw it now, but it for sure makes some fun gameplay

odecay: yup, I was thinking we could change the duration potentially or have it extend a bit if picked up fast enough.

src/item.rs
```diff
@@ -210,3 +215,156 @@ fn drop_system(
...
+
+/// A component that with Breakable, explodes.
+#[derive(Component, Clone)]
+pub struct Explodable {
+    pub attack: AttackMeta,
+    pub timer: Timer,
+    pub fusing: bool,
+    pub animated_sprite: AnimatedSpriteSheetBundle,
+    pub explosion_frames: AttackFrames,
+}
```
odecay, line 227: Does explodable need to have its own `AttackMeta` and `AttackFrames` ?
`AttackMeta` already has its own `AttackFrames` but couldnt `Explodable` be a component with only `timer` and `fusing` and `animated_sprite`?
Then add the `Explodable` component to an entity with an `AttackMeta` and query for both when you need to deal with both.

Zac8668: Not sure about the other thing you said, but we need two attack frames actually, one for the bomb explosion, and one for the frames the boss throws the bombs

src/attack.rs
```diff
@@ -108,6 +110,16 @@ fn activate_hitbox(
...
+        if let Ok((animation, explodable)) = explodable_query.get(**parent) {
+            if animation.current_frame >= attack_frames.startup
+                && animation.current_frame <= attack_frames.active
+            {
+                commands.entity(entity).insert(Collider::cuboid(
+                    explodable.attack.hitbox.size.x / 2.,
+                    explodable.attack.hitbox.size.y / 2.,
+                ));
+            }
+        }
```
odecay, line 122: I'd like to avoid special casing this activation code, I've done a bit of a refactor, added the hitbox metadata directly to `Attack` now which should let you remove this extra query and block if you add a `AvailableAttacks` component to the parent entity of the attack (im guessing the bomb entity in this case).

odecay (commented):
I think its good for now, maybe we can revisit after next release to see if we can simplify anything further

odecay (approved):

odecay:
bors merge

PR #273, opened 2022-11-15
Fix inactive sword hitbox bug, update docs, reflect

odecay:
Fixes a bug introduced when adding support for multiple attacks, adds in some reflection that was used while debugging, slightly simplified syntax for dealing with `AvailableAttacks` and added doc comment to specify its intended use.


odecay:
bors merge

PR #274, opened 2022-11-16
Attack hitbox insertion and activation refactor

odecay:
Changed some logic in hitbox activation system which should ensure it is correct for multiple attacks.
Added collider data directly to attacks on spawn, deduping attack code.


odecay:
bors merge

PR #277, opened 2022-11-20
Add ranged attack as Slinger fighter default attack

### LINKED ISSUE #268: Add ranged attack as Slinger fighter default attack.
### Description

- [ ] Communicate with artists about projectile sprite
- [ ] Add projectile attack state
- [ ] Update Slinger AI targeting to pick a target further from the player, and in smaller Y range from targeted player.
- [ ] Change Slinger default attack in YAML to projectile attack, and ensure it is used by its AI

### Alternatives & Prior Art

_No response_

Zac8668:
Closes #268


assets/fighters/slinger/slinger.fighter.yaml
```diff
@@ -45,18 +45,19 @@ spritesheet:
...
-  - name: "punch"
+  - name: "projectile"
     damage: 7
     frames:
-      startup: 1
-      active: 2
-      recovery: 3
+      startup: 4
+      active: 4
+      recovery: 4
```
odecay, line 55: having all these frames be the same *might* cause issues but maybe not. Will look more closely when I get a chance.

odecay:
Just pulled and tested, the arc and targeting is really cool!

There seems to be an issue with the targeting AI where sometimes they will turn around and fire backwards/away from the players.

Zac8668:
> There seems to be an issue with the targeting AI where sometimes they will turn around and fire backwards/away from the players.

Yeah, not sure why that happens, gonna try to fix it.

src/enemy_ai.rs
```diff
@@ -49,13 +53,43 @@ pub fn set_target_near_player(
...
+                            x_offset += throw_velocity.x
+                                * (lifetime * 0.65)
+                                * if p_transform.translation.x > e_transform.translation.x {
+                                    -1.
+                                } else {
+                                    1.
+                                };
+
+                            y_offset += (throw_velocity.y + gravity) * (lifetime * 0.65);
+                        }
```
odecay, line 91: I think this offset calculation is to blame for the slinger sometimes turning around when firing.
The offset was previously used for the melee attack only to add a bit of randomness to where the enemies would chose to stand and attack from, so they wouldnt all pile up directly on top of one another.

I think we should refactor a bit changing the `set_target_near_player` system to `set_move_target_near_player` to clarify intent, and move the projectile targeting code to its own system, which should be run when ranged enemy reaches their move target. likely need to introduce another component for projectile targeting.

assets/fighters/slinger/slinger.fighter.yaml
odecay: seems good the way its implemented, only the startup frame is ever being used :+1:

odecay (approved):

odecay:
bors merge

PR #284, opened 2022-12-05
sound working for chain attacks

odecay:
Updated assets for Dev and chain system to work with sound effects on each attack.


odecay:
bors merge

PR #288, opened 2022-12-30
upgrade to bevy 0.9

lecoqjacob:
This is from issue https://github.com/fishfolk/punchy/issues/287

Some wrapper types were needed to make bevy 0.9 resources happy.

Changed:
- EguiFontDefinitions that wraps egui::FontDefinitions
- LevelHandle - wraps Handle<LevelMeta>
- GameHandle - wraps Handle<GameMeta>
- Bevy rapier upgraded to using `Group` instead of base u32
- EguiRenderInputContainer - new wrapper type from bevy_egui for HashMap of keys

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ members = [
 ]
 
 [dependencies.bevy]
-version = "0.8"
+version = "0.9"
 default-features = false
 features = [
     "x11",
@@ -26,12 +26,12 @@ features = [
 punchy_macros = { path = "./macros" }
 
 anyhow = "1.0.58"
-bevy-parallax = "0.2.0"
-bevy_egui = "0.15.0"
-egui_extras = "0.18.0"
-bevy_kira_audio = { version = "0.11.0", features = ["mp3"] }
-bevy_rapier2d = { version = "0.16.0", features = ["debug-render"] }
-iyes_loopless = { version = "0.8.0", features = ["states"] }
+bevy-parallax = "0.3.0"
+bevy_egui = "0.18.0"
+egui_extras = "0.20.0"
+bevy_kira_audio = { version = "0.13.0", features = ["mp3"] }
+bevy_rapier2d = { version = "0.19.0", features = ["debug-render"] }
+iyes_loopless = { version = "0.9.0", features = ["states"] }
 serde = { version = "1.0.137", features = ["derive"] }
 serde_yaml = "0.9.2"
 thiserror = "1.0.31"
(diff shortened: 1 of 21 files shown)


src/main.rs
```diff
@@ -78,35 +73,44 @@ fn main() {
     let engine_config = &*config::ENGINE_CONFIG;
 
     let mut app = App::new();
-    app.insert_resource(WindowDescriptor {
-        title: "Fish Folk Punchy".to_string(),
-        scale_factor_override: Some(1.0),
-        ..default()
-    })
-    .insert_resource(ImageSettings::default_nearest());
-
```
odecay, line 87: we need a `ImagePlugin::default_nearest()` somewhere in the DefaultPlugins builder to replace this so the filtering mode is correct.

odecay (changes requested):

src/main.rs
lecoqjacob: Somehow this got lost in the upgrade. I’ll quickly add this back. Nice catch 😅

lecoqjacob:
Added `ImagePlugin` should be good now

zicklag:
You should be able to `cargo update` to get the latest version of `bevy_mod_js_scripting`, that should fix the Wasm CI issue.

odecay (approved):

Cargo.toml
```diff
@@ -1,70 +1,63 @@
...
+bevy_kira_audio = { version = "0.13.0", features = ["mp3"] }
+bevy_rapier2d   = { version = "0.19.0", features = ["debug-render"] }
+egui_extras     = "0.20.0"
+getrandom       = { version = "0.2", features = ["js"] }
+iyes_loopless   = { version = "0.9.0", features = ["states"] }
+rand            = "0.8.5"
+serde           = { version = "1.0.137", features = ["derive"] }
+serde_yaml      = "0.9.2"
+structopt       = "0.3.26"
+thiserror       = "1.0.31"
```
zicklag, line 32: Just curious, how did you format this file?

I think kind of like it, if there's an automatic way to do it.

lecoqjacob: Oh i apologize. that is my vs-code extensions `Even Better Toml` auto formatting my toml for me. I forget to disable it on other projects since some people hate formatters being run on their code

zicklag: If @odecay doesn't have a problem with it, I'm fine, it's no problem.

I was just curious what you used to format it, because I might want to use it. :)

odecay:
bors merge

Cargo.toml
odecay: I'm okay with it for now, if it becomes a problem to maintain formatting without that specific formatter in the future we can always change it back.

PR #292, opened 2023-01-07
Adjust Slinger ranged attack to remove Knockback

### LINKED ISSUE #290: Adjust Slinger ranged attack to remove Knockback
### Description

Change the knockback time and velocity for Slinger ranged attack to interrupt player for less time. Currently its very easy to become stunlocked by the Slingers projectile attacks.

Currently a thrown items attack velocity is determined by a const here https://github.com/fishfolk/punchy/blob/b6ff6a93311e6c49c055db518ce827cc24c75fa3/src/item.rs#L161 but this should be configurable per item.
This attack velocity is then used to determine HitStun velocity.

I also propose that we change the name of `Attack.velocity` and `HitStun.velocity` to `Attack.pushback` and `HitStun.pushback` to increase clarity of intended purpose.

Zac8668:
Closes #290


odecay (approved):
Looks good, made a small change

odecay:
bors merge

PR #299, opened 2023-01-24
Boss bomb throw always has another static bomb sprite behind the animated explosion

### LINKED ISSUE #298: Boss bomb throw always has another static bomb sprite behind the animated explosion
### Description

Looks like the bomb has an extra sprite for the throw that may not be removed/changed.

### To Reproduce

Activate the boss, look at the bombs as it throws and as they land on the ground before exploding.
A 2nd sprite behind the first is observable.

Zac8668:
Closes #298

### DIFF
--- a/src/fighter_state.rs
+++ b/src/fighter_state.rs
@@ -1447,7 +1447,9 @@ fn bomb_throw(
                             attack_enemy: false,
                         })
                         .insert(ItemBundle {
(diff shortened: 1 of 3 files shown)


odecay (commented):
I think code looks good, tested and it works as well.

odecay:
bors merge

PR #301, opened 2023-01-30
Changed hitstun to not transition state if duration is 0

odecay:
This means we can now have attacks which apply damage but do not interrupt other actions/state. Currently in use for the slingers rock throw attack.


odecay:
bors merge

